PR #98, opened 2013-04-17
Support for custom class mapping

dlorson:
You can specify a custom subclass that should be used to map the parsed data to a Ruby object, e.g. instantiate a SpecialPerson instead of Person based on the data hash `type` attribute:

```ruby
class Person
  include Her
  def self.class_for_data(data)
    data[:type] == "special" ? SpecialPerson : self
  end
end

class SpecialPerson
end
```

This is useful when mapping associations where the destination object(s) is based on a class hierarchy (e.g. STI)

Let me know if the style and spec coverage are sufficient.

### DIFF
--- a/lib/her/model/associations.rb
+++ b/lib/her/model/associations.rb
@@ -49,7 +49,7 @@ module Her
                 when :has_many
                   Her::Model::Attributes.initialize_collection(klass, :data => data[data_key])
                 when :has_one, :belongs_to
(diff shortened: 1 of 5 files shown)


hubert:
@romansergey It looks like the responsibility has been passed to you in to get this in shape for merging. I am on a mission to get to 0 open PRs, so am going to close this. Feel free to re-open if you have time to finish out the work.

Issue #114, opened 2013-05-04
Support for ActiveModel::Serializers defaults and http://jsonapi.org


prognostikos:
It seems that there is a project to standardize the format of JSON APIs [1](http://emberjs.com/blog/2013/05/03/ember-data-progress-update.html) [2](http://jsonapi.org/)

As noted in #75 Her does not currently support the default settings for e.g [ActiveModel::Serializers](https://github.com/rails-api/active_model_serializers) which also seems to be the basis for the format defined at jsonapi.org.

I would be happy to work on this if no one else is - just wanted to check first to avoid duplication.

groodt:
Whats the recommended approach for this now?

7h3kk1d:
In regards to singular resources. According to [http://jsonapi.org/format/#document-structure-individual-resource-representations](JSONAPI-spec) singular resources don't need ot be wrapped in an array. AMS using json_api format can also sends it this way.

zacharywelch:
Closing this issue due to inactivity although it looks like some progress on JSON API and AMS was made in the meantime. If anything specific is still missing in our latest [support](https://github.com/remiprev/her#activemodelserializers-support) for these two, feel free to reopen the issue or create a new one with more information.

Thanks!

PR #116, opened 2013-05-05
Support for ActiveModel::Serializers (WIP)

prognostikos:
Support for ActiveModel::Serializers defaults as discussed in #114

I've added `Her::Model.extract_array`.


fishermand46:
Any updates on this. The active_model_serializers feature is actually pretty broken without this being fixed. It spits out json that looks like this `{"users": [{"id": 2, "name": "John"}]}` but Her expects a root node for *each* object in the collection which is a format I have personally never even seen before.

Issue #97, opened 2013-04-16
Issue with inheritance


dsshap:
I found an issue with inheritance

``` ruby
module MyAPI
  class Model
    include Her::Model

    parse_root_in_json true
    include_root_in_json true
  end
end

class User < MyAPI::Model
end

User.find(1)
# this gives you
# GET /models/1
```

My work around is:

``` ruby
module foo
  class Bar
    def self.inherited(base)
      #example of things to inherit
      base.send(:include, Her::Model)
      base.send(:include, InstanceMethods)
      base.send(:uses_api, Foo.api)
      base.send(:primary_key, :_id)
    end

    module InstanceMethods
      def id
        self._id
      end
      def error?
        self.errors.any?
      end
    end
  end
end

module foo
  class User < Bar
  end
end

User.find(1)
# this gives you
# GET /users/1
```

PR #118, opened 2013-05-06
Fix for #97

prognostikos:
This was a wierd one, and I don't really like the fix but it does seem to work.

Check out the message for d32c742 for more details.


hubert:
discussion makes it seem like this is no longer an issue and no response for a long while.

PR #146, opened 2013-06-27
JRuby fix for associations

prcongithub:
In JRuby on Rails with Puma in production mode with config.threadsafe!, a Her::Model was getting loaded twice, so the associations in that model were added twice and there is no check for duplicate associations in Her::Model::Associations. This resulted in unexpected behavior when parsing has_many associations. association_class.parse gets invoked twice for the same association and second time the data argument passed contains Her::Association object instead of Hash. Hence it sets the attributes of association class instance as nil.

### DIFF
--- a/lib/her/model/associations.rb
+++ b/lib/her/model/associations.rb
@@ -31,7 +31,7 @@ module Her
         # @private
         def associations
           @_her_associations ||= begin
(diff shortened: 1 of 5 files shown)


bf4:
Close?

hubert:
i agree with @pencil assessment of the matter. @prcongithub if you have a solution that is able to address root cause, please reopen. thanks for the contribution!

PR #162, opened 2013-07-24
Allow specifying a custom finder key.

malclocke:
This allows support for cases where the API does not use the primary
key in the URL, for example /users/username rather than /users/1.

This has required a major rework of the request_path methods as the
correct path can no longer be inferred from the parameters alone.  The
decision as to whether to request a collection_path or resource_path has
been pushed up into each occurence of the calling code.

Closes #148

### DIFF
--- a/README.md
+++ b/README.md
@@ -674,6 +674,21 @@ user.destroy
 # DELETE "/users/4fd89a42ff204b03a905c535"
 ```
 
+### Custom finder key
+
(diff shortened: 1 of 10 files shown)


hubert:
@superp @malclocke is there anything here that is not addressed by the primary_key option?

FYI i'm trying to address all open PRs. if this is still an issue, please reopen and let me know. 

thanks for your contribution.

Issue #138, opened 2013-06-12
undefined method `full_messages' for {:name=>["can't be blank"]}:Hash


aaronmcadam:
I've set up my model according to this PR: https://github.com/remiprev/her/pull/101 and form template, but `full_messages` doesn't seem to be getting set.

See this gist: https://gist.github.com/aaronmcadam/5769512

PR #178, opened 2013-10-07
Add response errors to ActiveModel using ActiveModel::Errors

lleger:
**What**
This commit adds any response errors (such as those from calling `save`) to the `ActiveModel::Error` collection on the resource.

**Why**
Doing so enables accessing the errors hash via the `errors` attribute and allows the use of helper methods, such as `full_messages`. This also makes the `Her::Model` act more like an ActiveRecord object.

**How**
If the response errors are in an AR-like hash, then they're iterated over and added to the model via [`add`](http://api.rubyonrails.org/classes/ActiveModel/Errors.html#method-i-add) on `ActiveRecord::Error`. For example:

``` ruby
# from remote model
{ errors => @user.errors }.to_json

# parsed in @response_errors
{ :errors => { :name => ["should not be blank"] } }
```

This PR includes passing tests and has been integrated into a production app. This was briefly discussed in #138.

**Example**

``` ruby
class User
  include Her::Model
  # validates :name, presence: true (on remote model)
end

@user = User.new
@user.save

@user.errors # => { :name => ["can't be blank"] }
@user.errors.any? # => true
@user.errors.full_messages # => ["Name can't be blank"]
```

### DIFF
--- a/lib/her/errors.rb
+++ b/lib/her/errors.rb
@@ -19,7 +19,7 @@ module Her
       attr_reader :resource
       def initialize(resource)
         @resource = resource
-        errors = @resource.response_errors.join(", ")
+        errors = @resource.errors.full_messages.join(", ")
         super("Remote validation failed: #{errors}")
       end
     end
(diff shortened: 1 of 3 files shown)


noctivityinc:
@lleger I know this is an old issue but I completely agree with you.  Id like to try to integrate the same logic using a callback but unfortunately the callback is cancelled if an error occurs.  How did you eventually handle this?

lleger:
@noctivityinc To be honest, this is quite old. We abandoned Her not long after this in favor of greener pastures. Our fork has the code we ended up using, though, and you're free to use it if it's helpful. Good luck!

noctivityinc:
Thanks man.  What did you decided to use in place of Her for consuming
RESTful APIs in Rails, or are you using something else now on the front end
that is better suited?

Best,

Josh

lleger:
We ended up rewriting the app in question to use a React frontend that just straight consumed a Rails API. Much better experience than trying to consume an API on the backend and push back a response.

noctivityinc:
Thanks.  Ive been thinking of learning React for that exact purpose.  I
wrote the backend Rails API, which works great, but consuming it in a Rails
app with pseudo ActiveRecord code is proving to be *interesting*

PR #207, opened 2014-01-15
Save a reference to fetched has one relationships.

ryanto:
Currently has_one relationships are fetched everytime a message is sent
to the proxy. This commit adds a has_one cache that prevents the proxy
from fetching a new record if it has already fetched model with the
current params.

I added specs that show the behavior I am trying to accomplish. I think
the specs are straight forward, but if you run the against master they
will fail.


hubert:
@ryanto i'd prefer to the responsibility of caching be handled more generally at a lower level, e.g. the faraday_middleware referenced in the README, rather than sprinkling it in strategic places across the gem. 

feel free to reopen if you have a counterpoint you'd like to raise. thanks for contributing back!

PR #213, opened 2014-02-06
fix belongs to association load when using active model serializers

dvictory:
Belongs to does not work when using active model serializer format for models.


voltechs:
@remiprev can we get this merged in?

hubert:
@dvictory i believe that this issue was addressed in a recently merged PR. if you're still seeing the bug, please rebase with master and resubmit. 

thanks for contributing back!

Issue #231, opened 2014-04-30
Custom paths are broken


Chris911:
Hi,

It looks like the custom path feature doesn't work. I used the code in the README and this is what I get: 

``` Ruby
class User
  include Her::Model
  collection_path "/hello_users/:id"
end

u = User.find(1)
```

Result:

```
/her/lib/her/model/paths.rb:110:in `block in build_request_path': Missing :_id parameter to build the request path. Path is `/hello_users/:id/:id`. Parameters are `{}`. (Her::Errors::PathError)
    from /her/lib/her/model/paths.rb:107:in `gsub'
    from /her/lib/her/model/paths.rb:107:in `build_request_path'
    from /her/lib/her/model/relation.rb:93:in `block in find'
    from /her/lib/her/model/relation.rb:89:in `map'
    from /her/lib/her/model/relation.rb:89:in `find'
    from /her/lib/her/model/orm.rb:138:in `find'
    from test/test.rb:23:in `<main>'
```

Looks like something is wrong with that path:
 `Path is `/hello_users/:id/:id`. Parameters are `{}``

Since the error also mentions that `:_id` is missing (don't know why), I also tried with this line: `u = User.find(_id: 1)`. In that case, u is nil. 

If you have an idea of what could be wrong I'd like to go and fix it but I'm not sure where to start and when this was broken or if simply the documentation is out of date.

PR #232, opened 2014-05-01
Fix readme examples for custom paths

Chris911:
I was trying to get the custom paths working using the `collection_path` attribute. This feature looked broken (#231) but I think it's just the README that is out of date. I looked at [the code](https://github.com/remiprev/her/blob/master/lib/her/model/paths.rb#L41-L86) and found that `collection_path` should only change the base path with no attributes and that `resource_path` can take attributes. 

This PR changes the README examples by changing `collection_path` to `resource_path` where required and also adds an example for changing the base path using the `collection_path` attribute.

Also, I'm not so sure why [these tests are passing](https://github.com/remiprev/her/blob/master/spec/model/paths_spec.rb#L30-L47). With my test script and as pointed out in #231 I don't think they should work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,7 +411,7 @@ Resources must always have all the required attributes to build their complete p
 ```ruby
 class User
   include Her::Model
-  collection_path "organizations/:organization_id/users"
+  resource_path "organizations/:organization_id/users"
 end
 
 class Organization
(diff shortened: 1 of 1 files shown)


hubert:
hi @chris911 this may very well have been true at one point, but i don't think this is the case anymore. if you disagree, feel free to let me know.

thanks for contributing!

Issue #224, opened 2014-04-16
Creating nested resource POST's to wrong URL


danielstocks:
I try this (as per the documentation)

``` ruby
class User
  include Her::Model
  has_many :comments
  has_one :role
  belongs_to :organization
end

class Comment
  include Her::Model
end
```

When I run this code in the Rails console:

``` ruby
user = User.find(1)
user.comments.create()
```

The POST request is made to **/comments** and not **/user/1/comments.**

The documentation clearly states that the POST request should be done to /user/1/comments. Did I do something wrong here?

araslanov-e:
+1 (merge please)

nosretep:
Is this still a known/unsolved issue?

PR #238, opened 2014-05-05
Proposed solution to issue #224, getting the expected request_path when ...

ghost:
...creating a nested resource

This solution follows along the lines of what you do with the parent id and includes the parent request path when building the resource.

Please give some feedback on this, if it's a fair solution, if something is missing. I would like your thought on this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /pkg
 /tmp
 /coverage
+.idea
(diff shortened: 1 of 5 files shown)


chewi:
I'm not sure that this is such a good idea. Having an association doesn't necessarily mean that the parent path should be prepended like this. It really depends on the API. You can work around it by explicitly defining the collection_path but you may have noticed that this doesn't work for nested associations. I have fixed this in #318.

On the other hand, if you need to access a model directly as well as through an association then setting collection_path won't help. This is a tricky one. Maybe we need to be able to explicitly state the path that is used for direct access, if there is one, as well as the path used for each association. These paths could all be the same in some cases or they could all differ.

chewi:
@edtjones, if you're looking at this one, I'd really appreciate having a chat with you about it.

PR #262, opened 2014-06-27
Include has_one attributes correctly when calling to_params

dturnerTS:
Follow on to remiprev#206 to add support for has_one in to_params.

Also fixes remiprev#258

### DIFF
--- a/lib/her/model/parse.rb
+++ b/lib/her/model/parse.rb
@@ -61,7 +61,7 @@ module Her
         end
 
         def has_one_embedded_params(attributes)
-          present_has_ones = associations[:has_one].select { |a| attributes.include?(a[:data_key]) }
+          present_has_ones = (associations[:has_one] || []).select { |a| attributes.include?(a[:data_key]) }
           present_has_ones.compact.each_with_object({}) do |association, hash|
             params = attributes[association[:data_key]].to_params
             next if params.empty?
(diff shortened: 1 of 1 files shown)


filib:
@remiprev Besides this now causing conflicts, what was the reason this was never merged? This is causing pain for us and I'd really like to see this on `master`. How can I make that happen?

dturnerTS:
@filib I suspect it has to do with https://github.com/remiprev/her/issues/324 . Once things get back on track I'll be happy to merge master back into this PR

filib:
@dturnerTS Ah, didn't see that and didn't realise that @remiprev was looking for a new maintainer. Thanks for the heads-up!

hubert:
@dturnerTS i'm working to get Her down to 0 PRs. given there is already has_many serialization, i think it makes sense to do the same for has_one associations. 

few things before we merge:
1) how does it handle the case where the has_one association key is present but the value is nil? 
2) please rebase with master so that we can make sure this works with the current travis checks.
3) please remove the commit bumping the version.

dturnerTS:
@hubert Thanks for looking into this again. 
(1) I've changed the code to be more defensive. 
(2) Conflicts are resolved
(3) Version bump is removed

hubert:
@dturnerTS thanks for taking the time to go through this again. can you rebase this instead of merging it?

dturnerTS:
I'm not a big fan or rebasing. Can you explain why you think its preferable to the mater merge here?

Issue #215, opened 2014-02-20
Allow scopes on associations


forest:
I expected scopes on association to work, but they currently don't. Is this a planned feature, or something that is worth me investigating as an enhancement?

Any direction would be appreciated if this enhancement is desired for the project.

``` ruby
class Project
  include Her::Model

  has_many :iterations

  # This works.
  def current_iteration
    iterations.where(scope: :current).first
  end
end

class Iteration
  include Her::Model

  belongs_to :project
  has_many :stories

  scope :done, -> { where(scope: :done) }
  scope :current, -> { where(scope: :current) }
  scope :offset, ->(offset) { where(offset: offset) }
end

project = Project.find(12345)
iteration = project.current_iteration
```

This is what I want to be able to do, but `current` is undefined on the `Her::Model::Associations::HasManyAssociation`

```
iteration = project.iterations.current
iteration = project.iterations.done.offset(-5)
```

PR #268, opened 2014-07-03
Scopes on associations

dturnerTS:
Punch a hole in the association_proxy for scopes. Also inject the parents id if needed.

Addresses https://github.com/remiprev/her/issues/215

### DIFF
--- a/lib/her/model/associations/association.rb
+++ b/lib/her/model/associations/association.rb
@@ -20,7 +20,7 @@ module Her
           parent_id_string = "#{@parent.class.to_s.demodulize.downcase}_#{@parent.class.primary_key}"
           parent_id = @parent.send(@parent.class.primary_key)
           scoped = if klass.collection_path[parent_id_string]
-            klass.where(parent_id_string => parent_id)
+            klass.where("_#{parent_id_string}" => parent_id)
           else
             klass
           end


dturnerTS:
@hubert Anything I can do to help get this merged?

hubert:
@dturnerTS i've blocked out some time on saturday (tomorrow) to work through pending pull requests. 

thanks for the kind words. happy to contribute back to the gem.

chewi:
@edtjones, if you're looking at this one, I'd really appreciate having a chat with you about it.

edtjones:
@chewi definitely, and looks really useful. Need to absorb fully. Are there any downsides / expected failure modes which mean we should be careful about merging this in?

chewi:
I'd like it to work more like it does in Active Record, where you can call any class method through an association. The code I quoted above highlights the difference between Her and Active Record. I'm a little hazy on the details after all this time but Active Record is very mature now and I think we should strive to use the same approaches wherever possible as they have clearly served it well.

shashiprab:
Any plans on merging this PR?

edtjones:
hi @shashiprab @chewi this has been hanging around for ages - sorry! Don't have the mental bandwidth to deal with the implications of merging this at the moment but I would be delighted if someone else could review.

zacharywelch:
Remember happily applying this patch a few years back and eventually moving a few things around. Also have a rough patch that would let us remove `collection_path` as @kinopyo pointed out.

```ruby
class User
  include Her::Model
  has_many :comments
end

class Comment
  include Her::Model
  belongs_to :user
  scope :approved, -> { where(approved: true) }
end

> user = User.find(1)
# GET "/users/1"
> user.comments.approved
# GET "/users/1/comments?approved=true"
```

I'll put a PR together soon.

PR #269, opened 2014-07-07
Handle hash with string keys returned from faraday-http-cache

linhmtran168:
I'm using [faraday-http-cache](https://github.com/plataformatec/faraday-http-cache) with Redis but the parsed JSON hash return from the cache is not symbolised and caused Her to throw Nil errors. #192 does not fix all the problems.

### DIFF
--- a/lib/her/api.rb
+++ b/lib/her/api.rb
@@ -103,7 +103,7 @@ module Her
         end
       end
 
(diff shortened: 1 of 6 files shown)


gmontard:
Hello,

Any update on a merge of this PR?

Thanks.

Issue #248, opened 2014-05-23
subsequent calls when accesssing assosiation objects


revathskumar:
I have a `Store` model and `Inventory` model in which `Store` has_many `Inventories`.

The `her` document says 

> Subsequent calls to #comments, #role and #organization will not trigger extra HTTP requests and will return the cached objects.

But when I tied, `Her` is making subsequent calls.

``` ruby
@store = Store.find(1)
@products = @store.inventories # Fires first call to inventories which is ok.
@metadata = @products.metadata # Fires another request to inventories which is unexpected

@products. each do | product | # Again fires another request to inventories
    # do something with product
end
```

Any one facing the same issue?

I tried with `master` branch and `v0.6.8` same issue on both.

PR #277, opened 2014-08-08
Cache subsequent calls with parameters when accesssing assosiation objects

kinopyo:
### Changes

Change `@cached_result` type to a Hash, and use `@params` as hash key to store the value.

Based on this pull request: https://github.com/remiprev/her/issues/248

### Real World Example

When I apply pagination parameter to the association objects, I found that the request was not cached and will be triggered every time it's been called.

```ruby
@posts = @user.posts.where(page: params[:page])

@posts # => Fire a request
@posts # => Fire a request
@posts # => Fire a request
```

By using memcache the requests will still be sent, even though since 2nd time memcache will return the cached value.

---

Not sure if it has any undesirable side effect, I've checked that all specs are passed on my local.

### DIFF
--- a/lib/her/model/associations/association.rb
+++ b/lib/her/model/associations/association.rb
@@ -12,6 +12,7 @@ module Her
           @params = {}
 
           @klass = @parent.class.her_nearby_class(@opts[:class_name])
+          @cached_result = {}
           @name = @opts[:name]
         end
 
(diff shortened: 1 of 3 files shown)


kinopyo:
I'm no longer using this gem and have no intention on driving this pull request anymore so will just close it :)

PR #293, opened 2014-10-24
don't crash if errors is a hash

l8nite:
I was using Her with an API that returns a hash for errors instead of an array:

```
[2] pry(#<Her::Errors::ResourceInvalid>)> @resource.response_errors
=> {:param1=>["can't be blank"], :base=>["duplicate enrollment pending"]}
```

lib/her/errors.rb crashed on .join() being undefined

### DIFF
--- a/lib/her/errors.rb
+++ b/lib/her/errors.rb
@@ -19,7 +19,14 @@ module Her
       attr_reader :resource
       def initialize(resource)
         @resource = resource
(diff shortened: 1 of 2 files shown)


l8nite:
@hubert is this one still worth merging?

hubert:
@l8nite i think this responsibility falls to the parser setting up the data, meta, and errors in the expected way.

PR #294, opened 2014-10-30
Wrap parsed data which maintains backwards compatibility for apis that r...

hubert:
...eturn individual resources as one element arrays

Per json api spec, an *"individual resource SHOULD be represented as a single 'resource object'"*, not a one-element array. From their documentation

```
{
  "posts": {
    "id": "1",
    // ... attributes of this post
  }
}
```

this change maintains backwards compatibility, but updates specs and documentation to espouse the recommended specification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -594,7 +594,7 @@ class User
 end
 
 user = Users.find(1)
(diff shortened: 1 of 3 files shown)


hubert:
closing this. json api support will be handled as a first-class representation.

PR #301, opened 2014-11-17
Handle has one and accepts nil value for it

pisaruk:
Now has_one associations will be taken into account when calling to_params in the parent model. 
That allow us to save a parent model containing has_one associations.
For example:

    class Person
      include Her::Model
      has_one :nick_name
      attributes :name
    end

    class NickName
      include Her::Model
      attributes :origin, :author, :created_at
    end

    p = Person.first
    p.assign_attributes(name: 'José', 
      nick_name_attributes: {origin: 'high_school', author: 'Big Guy', created_at: 18.years.ago})
    p.save

Besides that nil value will be accepted:

    p = Person.first
    p.nick_name = nil
    p.save

### DIFF
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -39,7 +39,6 @@ module Her
 
         run_callbacks callback do
           run_callbacks :save do
-            params = to_params
             self.class.request(to_params.merge(:_method => method, :_path => request_path)) do |parsed_data, response|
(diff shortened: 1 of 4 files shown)


diegoy:
This feature is useful, is there any problem in this implementation in particular?

hubert:
most of the code here is PRd in #262 which i've responded to the OP about. going to close this for now and wait for a few small changes before we try to merge it.

thanks for contributing!

PR #303, opened 2014-11-28
fix: use a lambda to define default her api

gottfrois:
Allowing to setup the api after models are loaded.


hubert:
hi @gottfrois if you rebase with master to get the latest checks and resubmit without the commit/revert, i'll merge this.

thanks for the contribution!

PR #314, opened 2015-01-13
Double quotes required for interpolation.

caomania:
https://github.com/remiprev/her/commit/5020082333c7faa6362060da85b6a3a91e6d6db2#diff-7280373369c4cbefb4c1e7c1df8b8a36R142 Introduced a bug which affects the ActiveModel::Dirty implentation. Double quotes should be used for interpolation!

This bug causes any attribute modification to be tracked as a change, even when the "new" value is the same as the existing value.

### DIFF
--- a/lib/her/model/attributes.rb
+++ b/lib/her/model/attributes.rb
@@ -192,7 +192,7 @@ module Her
             unless instance_methods.include?(:"#{attribute}=")
               define_method("#{attribute}=") do |value|
                 @attributes[:"#{attribute}"] = nil unless @attributes.include?(:"#{attribute}")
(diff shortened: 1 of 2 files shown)


mtwentyman:
I've run into this myself and had to patch this method over.  +1

paulvt:
Maybe also add a test?

mtwentyman:
The test I've added here will fail without the corresponding patch.

mtwentyman:
@hubert would you mind taking a quick look at this one?  this fixes a regression as noted in the initial comment.

hubert:
merged. thanks for contributing. :+1:

PR #317, opened 2015-02-11
fix: make response parsing comply with JSON api for individual objects

hderms:
In [this](http://jsonapi.org/format/#document-structure-individual-resource-representations) section of the JSON api spec, we see that it is possible for a JSON api spec compliant service to return a response in the form of an 'individual resource' object.
Example:
Both the following are valid
``` JSON
 {"users" : [{"id" : 1}]}
```
``` JSON
 {"users" : {"id" : 1}}
```

This fix makes it handle the "collection object" case properly. 
This behavior was mentioned in issue #114

### DIFF
--- a/lib/her/model/attributes.rb
+++ b/lib/her/model/attributes.rb
@@ -172,7 +172,8 @@ module Her
         # @private
         def new_from_parsed_data(parsed_data)
           parsed_data = parsed_data.with_indifferent_access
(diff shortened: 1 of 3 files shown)


7h3kk1d:
:+1:

aahmad:
:100: - follows spec

hubert:
@hderms there is more first class support for jsonapi in the latest versions of the gem, which follow the jsonapi 1.0 spec. 

closing now. thanks for the contribution!

PR #321, opened 2015-03-09
avoid blank hash in embeded parameter

y00rb:
Hi Guys
just improve the generate embedded attributes(has_many relation).
if the has_many value is blank, will make errors in `embeded_params`, such as `TypeError: no implicit conversion of nil into Hash`, `NoMethodError: undefined method `[]=' for nil:NilClass`

so I have add method to make default in inject, set blank hash.


y00rb:
@remiprev  Could you checkout it? if it is merge, I can used new version in my project

PR #325, opened 2015-03-13
Return result from `save` when using `create`

jeromegn:
`create` and `save` should behave the same (`create` being a shortcut for instantiation + saving). This is useful because `save` can sometimes return false.


jeromegn:
Actually, it looks like that's what Rails does. My PR would make Her not compliant with Rails.

Initially, I created this because if our API returned an error status, there would be no way of knowing that it failed (expect that there wouldn't be an `id` on the element)

Closing!

PR #329, opened 2015-03-31
Update association_proxy.rb - singleton class can't be dumped error while trying to her model obejct in Rails.cache (file_store))

crv:
The metaclass defined here is causing the problem when I try to store her::Model object in Rails.cache with file_store. See the example below. Because if this I am not able to use latest her version in my project. Rails cache works fine with her version 0.6.8 after that it gives following error. 

```
[2] first »  metaclass = (class << self; self; end) 
=> #<Class:#<Object:0x000000017ce3f8>> < Object
[3] first »  Marshal.dump(metaclass)
TypeError: singleton class can't be dumped
from (pry):3:in `dump'
```

### DIFF
--- a/lib/her/model/associations/association_proxy.rb
+++ b/lib/her/model/associations/association_proxy.rb
@@ -33,8 +33,7 @@ module Her
           end
 
           # create a proxy to the fetched object's method
-          metaclass = (class << self; self; end)
-          metaclass.install_proxy_methods 'association.fetch', name
+          AssociationProxy.install_proxy_methods 'association.fetch', name
 
           # resend message to fetched object
           __send__(name, *args, &block)
--- a/spec/model_spec.rb
+++ b/spec/model_spec.rb
@@ -22,6 +22,10 @@ describe Her::Model do
     it { should have_key(:name) }
     it { should have_key(:comments) }
   end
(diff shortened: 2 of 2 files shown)


crv:
Can someone please review this and if found okay, please merge.

cknoxrun:
+1

ff-cviradiya:
@remiprev @pencil 
Any review comments, suggestions ? Please share if any.

ff-rboutotte:
Con someone accept this pull request? Or explain why you can't accept the pull request? We have same issue and this has existed for quite some time, and patch provided that works.

hubert:
@crv couple things would be needed to get this merged in. apologies for the long wait.

1) could you update the test so that it asserts that whatever is dumped can also be loaded? i'm dubious of tests that only check that an error has not occurred.
2) this needs to get rebased with what's in master so that it runs the latest travis checks and we can be sure it works with the latest.

crv:
Hi @hubert ,

I have created new pull https://github.com/remiprev/her/pull/377 which has latest from master and a new spec as you suggested. Please review and merge if it looks proper.

hubert:
closed in favor of #377

PR #333, opened 2015-04-21
Typo in documentation

fist28:
Method `request` isn't private.


hubert:
@fist28 my read of the code is that the method is not private in the ruby sense, but rather part of the "private" api of her. 

thanks for the contibution!

PR #335, opened 2015-04-30
Prevent syntax errors when `name` is a Symbol for a setter.

gdeoliveira:
I've experienced an issue when trying to perform an assignment on an associated record. It complained about a syntax error on the modified line (specifically the `&` character). This change seems to solve that issue.

### DIFF
--- a/lib/her/model/associations/association_proxy.rb
+++ b/lib/her/model/associations/association_proxy.rb
@@ -8,7 +8,7 @@ module Her
           names.each do |name|
             module_eval <<-RUBY, __FILE__, __LINE__ + 1
               def #{name}(*args, &block)
-                #{target_name}.#{name}(*args, &block)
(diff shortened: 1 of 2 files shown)


hubert:
could you add a test to illustrate the fixed behavior?

gdeoliveira:
I've modified an existing test that will fail with the original implementation but will pass with the modification (under 2.0.0-p645). Please bear in mind that with both the old and new implementation tests will fail under the latest Ruby version (2.2.2) with the following:
```
NoMethodError:
  private method `__method__' called for ...
```
This can be fixed by changing `public_send` to `send` in the line I modified, but perhaps you'd like to modify `Her::Model::Attributes#method_missing` instead.

gdeoliveira:
This has already been fixed at master.

PR #307, opened 2014-12-10
Memoize user-defined methods during Her attribute definition

toddmazierski:
585e91c fixed a bug where attributes defined by Her would overwrite user-defined methods, however, it also appears to have introduced a performance regression. According to the profiler tool [ruby-prof](https://github.com/ruby-prof/ruby-prof), when initializing new objects, `Module#instance_methods` takes up a significant chunk of time (observed with Ruby 1.9.3 and 2.1.5).

If it's safe to assume that it's unlikely the set of user-defined methods will change *while* this method is being evaluated, memoizing them can result in a performance boost.

### Before

```ruby
Benchmark.realtime { User.all.count }
# => 1.083330651
```

#### ruby-prof report

```
Thread ID: 70011050636080
Fiber ID: 70011059749060
Total: 0.987885
Sort by: self_time

 %self      total      self      wait     child     calls  name
 40.69      0.402     0.402     0.000     0.000     6610   Module#instance_methods
  5.14      0.051     0.051     0.000     0.000    64530   Symbol#to_s
  3.82      0.057     0.038     0.000     0.019    19833   String#%
  2.95      0.057     0.029     0.000     0.028    13702   Array#include?
  2.43      0.050     0.024     0.000     0.026    19830   ActiveModel::AttributeMethods::ClassMethods#instance_method_already_implemented?
  2.39      0.036     0.024     0.000     0.012    19530   ActiveSupport::HashWithIndifferentAccess#convert_key
  2.15      0.021     0.021     0.000     0.000    67642   Symbol#==
  2.03      0.069     0.020     0.000     0.049     9579   ActiveSupport::HashWithIndifferentAccess#[]=
  1.92      0.028     0.019     0.000     0.009    10287  *ActiveSupport::HashWithIndifferentAccess#convert_value
  1.85      0.075     0.018     0.000     0.057    19830   ActiveModel::AttributeMethods::ClassMethods::AttributeMethodMatcher#method_name
```

### After

```ruby
Benchmark.realtime { User.all.count }
# => 0.695653897
```

#### ruby-prof report

```
Thread ID: 69820109198100
Fiber ID: 69820121752620
Total: 0.606979
Sort by: self_time

 %self      total      self      wait     child     calls  name
  5.77      0.045     0.035     0.000     0.010    19833   String#%
  5.35      0.032     0.032     0.000     0.000    64670   Symbol#to_s
  4.92      0.059     0.030     0.000     0.029    13719   Array#include?
[...]

PR #336, opened 2015-05-01
use a set rather than array to test existance of instance methods

faisalmansoor:
Her::Model::Attributes#attributes searches an array in a loop which causes a lot of performance overhead in large object graphs. 

Using a set rather than an array for lookup should significantly improve performance.

### DIFF
--- a/lib/her/model/attributes.rb
+++ b/lib/her/model/attributes.rb
@@ -186,10 +186,12 @@ module Her
         def attributes(*attributes)
           define_attribute_methods attributes
 
+          instance_methods_set = Set.new instance_methods
+
           attributes.each do |attribute|
             attribute = attribute.to_sym
 
-            unless instance_methods.include?(:"#{attribute}=")
+            unless instance_methods_set.include?(:"#{attribute}=")
               define_method("#{attribute}=") do |value|
                 @attributes[:"#{attribute}"] = nil unless @attributes.include?(:"#{attribute}")
                 self.send(:"#{attribute}_will_change!") if @attributes[:"#{attribute}"] != value
@@ -197,7 +199,7 @@ module Her
               end
             end
 
-            unless instance_methods.include?(:"#{attribute}?")
+            unless instance_methods_set.include?(:"#{attribute}?")
               define_method("#{attribute}?") do
                 @attributes.include?(:"#{attribute}") && @attributes[:"#{attribute}"].present?
               end


hubert:
hi faisal. how does using a set, rather than an array, improve search performance? it also takes O(n) time to find something in a set.

marshall-lee:
`Set` is implemented using hash tables so it should  be `O(1)`. But on small sets it may not improve performance due to overhead.
However, this one improves performance due to memoization — current implementation calls `instance_methods` which creates a new array on each iteration so it's slow.

But it's just a variation of #307 and you can see benchmarks — they are adequate. We faced this problem too — current implementation parses json in our project for about *10 seconds* but that patch dramatically increased the performance!

Also please look at my #337 — it solves this problem even better without creating any additional arrays at all.

marshall-lee:
I benchmarked this PR in https://github.com/remiprev/her/pull/337#issuecomment-99233830

faisalmansoor:
Searching a set can take O(1) to O(lg n) time, depending upon how its implemented, not sure how its implemented in ruby. method_defined is implemented using a binary search which takes O(lg n), but, since it does not build a separate data structure it might run faster in practice.

Both solutions should work a lot faster than what we have right now, so, we can choose whichever has better benchmark.  
 
```
static VALUE
rb_mod_method_defined(VALUE mod, VALUE mid)
{
    if (!rb_method_boundp(mod, rb_to_id(mid), 1)) {
        return Qfalse;
    }
    return Qtrue;

}
```

faisalmansoor:
#337 is a simpler solution and does not require the overhead of creating a Set, so we should use that.

marshall-lee:
> method_defined is implemented using a binary search which takes O(lg n)

:smiley: despite the name `rb_method_boundp` it's not a binary search. It's a hash lookup.

Just trace the [https://github.com/ruby/ruby/blob/ruby_2_2/vm_method.c](source code):

 * `rb_method_boundp` calls `rb_method_entry_without_refinements`
 * `rb_method_entry_without_refinements` calls `rb_method_entry`
 * `rb_method_entry` calls `rb_method_entry_get_without_cache`
 * `rb_method_entry_get_without_cache` calls `search_method`
 * **(the most interesting)** `search_method` calls `lookup_method_table` *for each* superclass of a class (beginning from the class itself of course)
 * `lookup_method_table` calls `st_lookup` from [st.c](https://github.com/ruby/ruby/blob/ruby_2_2/st.c) which is a hash table lookup

hash lookup is O(1), so the overall complexity is just O(M) where M is the number of superclasses of a class.

Interesting: looking up in M small hash tables is turned out to be better than looking in one big hash table (the constructed `Set`). I still think this is because of `Set` constructing overhead + maybe some caching techniques deep inside  `method_defined?` call.

marshall-lee:
> Searching a set can take O(1) to O(lg n) time, depending upon how its implemented

Ruby's `Set` is not ordered by value so It's not a balanced search tree :smiley: (but ordered by the order of inserts by the way — just like `Hash`)
Look at implementation of [set.rb](https://github.com/ruby/ruby/blob/ruby_2_2/lib/set.rb) — it just piggybacks to `Hash.new`.

marshall-lee:
By the way, method lookup in Rubinius is implemented the similar way — a hash lookup for each superclass.

https://github.com/rubinius/rubinius/blob/2.2/kernel/common/module.rb:

 * `method_defined?` calls `lookup_method`
 * `lookup_method` calls `mod.method_table.lookup` for each superclass `mod`.
 * `method_table.lookup` is defined in [vm/builtin/method_table.cpp](https://github.com/rubinius/rubinius/blob/2.2/vm/builtin/method_table.cpp) which is a typical hash table implementation.

faisalmansoor:
@marshall-lee thanks for the details, very informative, thanks. 
Faisal

PR #338, opened 2015-05-07
:attributes class method misbehavior

marshall-lee:
In #337 a following case was [discussed](https://github.com/remiprev/her/pull/337#issuecomment-99238263):

```ruby
class A
  attr_accessor :foo
end

class B < A
  include Her::Model

  attributes :foo
end
```

One would expect that `:foo`, `:foo=` methods would be overriden but current implementation behaves not as expected. To be precise, `:foo` will be overriden but `:foo=` will not. Why? Because Her defines `:foo` and `:foo=` differently. Reader method `:foo` is defined using `define_attribute_method` from ActiveModel but `:foo=` is defined manually (but only if a method doesn't exist already). ActiveModel's `define_attribute_method` checks a method to exist too but it's doing it smarter. It doesn't call `method_defined?` for a whole class — it checks `generated_attribute_methods` which is a dynamically included module:

```ruby
def generated_attribute_methods #:nodoc:
  @generated_attribute_methods ||= Module.new {
    extend Mutex_m
  }.tap { |mod| include mod }
end

# ...

def instance_method_already_implemented?(method_name) #:nodoc:
  generated_attribute_methods.method_defined?(method_name)
end
```

So, ActiveModel is behaving smarter — it checks method for extistence but *only among the methods defined by itself*. ActiveModel has this behavior in 3.x and 4.x and we definitely should make Her to comply it. This issue is definitely a bug and we cannot wait for 1.0.0 to fix it.

This issue is test first so I could try to make it green.

### DIFF
--- a/lib/her/model.rb
+++ b/lib/her/model.rb
@@ -67,6 +67,9 @@ module Her
       # Configure ActiveModel callbacks
       extend ActiveModel::Callbacks
       define_model_callbacks :create, :update, :save, :find, :destroy, :initialize
+
+      # Define matchers for attr? and attr= methods
+      define_attribute_method_matchers
     end
   end
 end
--- a/lib/her/model/attributes.rb
+++ b/lib/her/model/attributes.rb
@@ -184,20 +184,22 @@ module Her
         #     attributes :name, :email
         #   end
         def attributes(*attributes)
-          define_attribute_methods attributes
-
           attributes.each do |attribute|
-            unless method_defined?(:"#{attribute}=")
-              define_method("#{attribute}=") do |value|
-                @attributes[:"#{attribute}"] = nil unless @attributes.include?(:"#{attribute}")
-                self.send(:"#{attribute}_will_change!") if @attributes[:"#{attribute}"] != value
-                @attributes[:"#{attribute}"] = value
-              end
-            end
+            define_attribute_method attribute
 
-            unless method_defined?(:"#{attribute}?")
-              define_method("#{attribute}?") do
-                @attributes.include?(:"#{attribute}") && @attributes[:"#{attribute}"].present?
+            generated_attribute_methods.module_eval do
+              unless method_defined?(:"#{attribute}=")
+                define_method("#{attribute}=") do |value|
+                  @attributes[attribute] = nil unless @attributes.include?(attribute)
+                  self.send(:"#{attribute}_will_change!") if @attributes[attribute] != value
+                  @attributes[attribute] = value
+                end
+              end
+
+              unless method_defined?(:"#{attribute}?")
+                define_method("#{attribute}?") do
+                  @attributes.include?(attribute) && @attributes[attribute].present?
+                end
               end
             end
           end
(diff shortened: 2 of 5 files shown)


marshall-lee:
https://travis-ci.org/remiprev/her/jobs/61596212 — it fails as expected.

But only 4 of 6 new spec examples failed. Those that test getter method (which is defined by ActiveModel) are green. So we should fix a behavior of setter and predicate methods.

marshall-lee:
We have two ways to fix it:

- Piggyback on ActiveModel's `generated_method_attributes`.
- Dynamically include our own module which stores setters and predicates. This will allow us to not depend on ActiveModel's implementation.

hubert:
nice investigation. 

my inclination would be to piggyback on activemodel since Her::Model does already for a number of features. given that activemodel provides a standard for behavior that is well-known, adhering to it would be intuitive for developers.

if there are counterargs, i'd be interested to hear them.

marshall-lee:
No counterargs from me too. So lets proceed.

Chosing this way we should decide what to do with concurrency. ActiveModel 4.x has a mutex on `generated_attribute_methods` that's being acquiring by ActiveRecord with `synchronize`. ActiveModel 3.x has no such mechanizm. But my opinion is that we should have such mutex anyway.

marshall-lee:
Then the question is: why we have `>= 3.0.0` in gemspec but only `3.2` in `travis.yml`? What versions of ActiveModel we actually support?

marshall-lee:
I tried this mutex workaround and it's ok:

```ruby
# Create a mutex for dynamically generated attribute methods
#
# @private
def attribute_methods_mutex
  @attribute_methods_mutex ||= if generated_attribute_methods.respond_to? :synchronize
                                 generated_attribute_methods
                               else
                                 Mutex.new
                               end
end
```

marshall-lee:
Last commit removes `method_defined?` workaround at all because ActiveModel already has a convention for defining custom attribute methods that our setter and predicate methods actually are. For instance, ActiveRecord uses `attribute_method_suffix`  for [predicate](https://github.com/rails/rails/blob/1492e83e49f74b1d874585000bad5368d9f09218/activerecord/lib/active_record/attribute_methods/query.rb#L7) and [setter](https://github.com/rails/rails/blob/02a48fb83c6aff2c9fc82e962286212ba26a91c5/activerecord/lib/active_record/attribute_methods/write.rb#L20) too.

marshall-lee:
cee317f acquires a mutex around an attribute methods definition and the specs for this are written.

Also I measured a performance as in #337 — it's not degraded but I think it's even became faster a bit.

marshall-lee:
@hubert ?

marshall-lee:
@hubert I rebased against current master (with your work on JSON API support). There were some conflicts in `spawn_model` spec macro, so I decided to change it this way: d4eec52b6c115973fb8f6e1fda4461bd636e3aaf.

Please merge it. This PR finally fixes the issue with attributes.

PR #341, opened 2015-05-14
Remove usused code

aderyabin:
Attribute `base_uri` is unused and probably it came from legacy


aderyabin:
//cc @hubert

marshall-lee:
In this file I also don't like that `Her::API.setup` calls instance method `#setup` twice — inside `#initialize` and also explicitly in `.setup`. Create a separate PR for this?

marshall-lee:
Thanks! Merging this PR probably helped me with https://github.com/remiprev/her/pull/351

PR #342, opened 2015-05-14
Don't call API#setup twice

marshall-lee:
Self-descriptive. No need to call `#setup` because `#initialize` already calls it.


marshall-lee:
@hubert ?

hubert:
hi @marshall-lee, i know you didn't introduce this issue, but how do you feel about removing the 'self' part from the initialize method. at a glance it's misleading when you're tracing the code from setup to initialize. 

otherwise G2G.

marshall-lee:
@hubert done!

hubert:
beautiful @marshall-lee 

thanks!

PR #343, opened 2015-05-15
Non-persisted records try to make HTTP calls for associations with nil foreign_key

valo:
If you create a new instance of a Her model and it has a belongs_to association which is nil it will try to fetch a record with ID=nil.

A short example:

```ruby
user = User.new(name: "Karlin", organization_id: nil)
user.organization # this throws an exception as it tries to fetch GET /organizations
```

The error from the test in the PR:

```
  1) Her::Model::Associations handling associations without details returns nil if the foreign key is nil
     Failure/Error: @user_without_organization_and_not_persisted.organization.should be_nil
     Faraday::Adapter::Test::Stubs::NotFound:
       no stubbed request for get /organizations
     # ./lib/her/api.rb:94:in `request'
     # ./lib/her/model/http.rb:56:in `request'
     # ./lib/her/model/http.rb:81:in `get_raw'
     # ./lib/her/model/http.rb:70:in `get'
     # ./lib/her/model/associations/belongs_to_association.rb:84:in `fetch'
     # ./lib/her/model/associations/association_proxy.rb:11:in `nil?'
     # ./lib/her/model/associations/association_proxy.rb:40:in `method_missing'
     # ./spec/model/associations_spec.rb:215:in `block (3 levels) in <top (required)>'
```

I think the correct behavior is to return nil no matter if the record is persisted or not.

### DIFF
--- a/lib/her/model/associations/belongs_to_association.rb
+++ b/lib/her/model/associations/belongs_to_association.rb
@@ -74,7 +74,7 @@ module Her
         def fetch
           foreign_key_value = @parent.attributes[@opts[:foreign_key].to_sym]
           data_key_value = @parent.attributes[@opts[:data_key].to_sym]
-          return @opts[:default].try(:dup) if (@parent.attributes.include?(@name) && @parent.attributes[@name].nil? && @params.empty?) || (@parent.persisted? && foreign_key_value.blank? && data_key_value.blank?)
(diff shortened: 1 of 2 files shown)


hubert:
@valo could you rebase with what's in master so we can run the full travisci checks. once that's done, we're G2G.

valo:
Done

hubert:
thanks for contributing!

PR #266, opened 2014-07-03
Fix associations with active model serializers

minktom:
Fix active model serializers format with associations (Issue #212)

PR #344, opened 2015-05-15
Fix belongs_to associations for models with active_model_serializers format

valo:
If you have two associated models like this:

```ruby
class User
  parse_root_in_json true, :format => :active_model_serializers
  belongs_to :organization
end

class Organization
  parse_root_in_json true, :format => :active_model_serializers
end
```

And you try to do:

```ruby
user = User.find(1)
user.organization
```

you get an exception:

     NoMethodError:
       undefined method `keys' for nil:NilClass
     # ./lib/her/model/parse.rb:142:in `root_element_included?'
     # ./lib/her/model/parse.rb:22:in `parse'
     # ./lib/her/model/attributes.rb:175:in `new_from_parsed_data'
     # ./lib/her/model/associations/belongs_to_association.rb:85:in `block in fetch'
     # ./lib/her/model/associations/belongs_to_association.rb:84:in `tap'
     # ./lib/her/model/associations/belongs_to_association.rb:84:in `fetch'
     # ./lib/her/model/associations/association_proxy.rb:11:in `id'
     # ./lib/her/model/associations/association_proxy.rb:40:in `method_missing'

I added a failing test for this case and implemented a fix, where it is used `get_raw` instead of `get` when fetching a belongs_to association. This is needed because `get` assumes the data to be parsed is a collection if you have format: active_model_serializers, which is not correct. There is just one object in a belongs_to association.

### DIFF
--- a/lib/her/model/associations/belongs_to_association.rb
+++ b/lib/her/model/associations/belongs_to_association.rb
@@ -81,8 +81,10 @@ module Her
 
           path_params = @parent.attributes.merge(@params.merge(@klass.primary_key => foreign_key_value))
           path = build_association_path lambda { @klass.build_request_path(path_params) }
-          @klass.get(path, @params).tap do |result|
-            @cached_result = result if @params.blank?
+          @klass.get_raw(path, @params) do |parsed_data, response|
+            @klass.new_from_parsed_data(parsed_data).tap do |result|
+              @cached_result = result if @params.blank?
+            end
           end
         end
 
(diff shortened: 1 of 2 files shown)


7h3kk1d:
Does this fix `https://github.com/remiprev/her/pull/266`?

valo:
#266 seems to fix a problem with `has_many` relations and parsing included data, like `user: { comments: [...] }` I merged the specs from #266 into my branch and there are failing specs for the included data case and there is a failing spec when using `user.comments.find`.

My PR is fixing only the `belongs_to` case.

hubert:
i merged in #359 which i believe should address this as well. feel free to re-open if you are still having issues. 

thanks for the contribution!

Issue #345, opened 2015-05-15
Her JSON API support


hubert:
Hi All. There's been great activity with issues and PRs lately. thank you.

One of the things, I'm keen to see with Her moving forward is first class support for [JSON API](http://jsonapi.org/). It's got great momentum behind it and addresses a lot of the issues we deal with building APIs. 

I propose we introduce a module such that the way you specify your model consumes and produces JSON API is a one-liner.

```ruby
class Contributor
  include Her::JsonApi::Model

  # defaults to demodulized, pluralized class name, e.g. contributors
  type :developers
end
```

JSON API response follows this general format:

```ruby
{ data: {
  type: "developers",
  id: "6ab79c8c-ec5a-4426-ad38-8763bbede5a7",
  attributes: {
    language: "ruby",
    name: "avdi grimm",
  }
}
```

Underneath the hood, we introduce a parsing and serialization routine that translates to and from the JSON API format and the format Her expects. 

Here are some benefits of this approach:
* Developer interaction with Her::JsonApi::Model would remain consistent with Her::Model, which should be easy for Rails developers to grok, since it's based on ActiveModel. 
* This solution builds off of, rather than modifies, the existing codebase. This makes it a low risk addition which should have no impact on existing functionality. 

The main downside is that we would need to combine the top-level data namespace with the attributes namespace. If an "id" or "type" were passed in attributes different than the top level id, we'd need to figure out a way to resolve the conflict. Perhaps top-level "id" and "type" get added as "data_id" and "data_type" or something...

marshall-lee:
1. What's the purpose of `format: :json_api` then? Deprecate it?
2. Maybe just create a PR so we can look at it?

marshall-lee:
> The main downside is that we would need to combine the top-level data namespace with the attributes namespace. If an "id" or "type" were passed in attributes different than the top level id, we'd need to figure out a way to resolve the conflict.

This issue should definitely be resolved before merging.
First of all: what JSON API standard says about issues like this? Probably It says nothing because there is actually no such problem on JSON API side :smiley: 

> Perhaps top-level "id" and "type" get added as "data_id" and "data_type" or something...

wow, just don't do this! What if I have `data_type` or `data_id` in attributes? Just don't put in ActiveModel's attributes something that is actually not an attribute because it will produce **bugs**.

[...]

PR #347, opened 2015-06-12
Initial jsonapi support (without compound documents)

hubert:
This is a first step in providing first-class support for JSON API within Her. It targets the 1.0 spec. Currently, it supports the basic CRUD operations and metadata, but does not yet have association support baked in. Based on what I have seen so far, that should be doable to add. See #345 for some additional background.

The primary design criteria here was to provide JSON API support without making any changes to existing functionality so as not to risk breaking anything already in place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -586,20 +586,44 @@ users = Users.all
 
 #### JSON API support
 
-If the API returns data in the [JSON API format](http://jsonapi.org/) you need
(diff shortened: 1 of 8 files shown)


Fivell:
:+1:

senid231:
+1 
very good idea

PR #354, opened 2015-07-08
Escape path variables

marshall-lee:
When dealing with parameters in a query string (`?a=1&b=blabla`) Faraday escapes keys and values using `Faraday::Utils.escape` helper. But when parameters are passed in a path string (RESTful way) it doesn't happen so for example when `id` value is something non-ascii.
Since http path building is done on the Her side, it should be fixed here.

### DIFF
--- a/lib/her/model/paths.rb
+++ b/lib/her/model/paths.rb
@@ -106,8 +106,13 @@ module Her
 
           path.gsub(/:([\w_]+)/) do
             # Look for :key or :_key, otherwise raise an exception
-            value = $1.to_sym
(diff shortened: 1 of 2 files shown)


marshall-lee:
This breaks our api clients on some ids :sob:

hubert:
thanks for the contribution. couple questions:

1) should we be escaping the primary key value as well?

2) could you update the test so that it is testing the public interface for path building (request_path) vs the underlying class method?

marshall-lee:
@hubert it escapes primary key as you may see in the new add spec.

marshall-lee:
@hubert wait, you mean rewrite all the tests to use `request_path` ?

marshall-lee:
updated

hubert:
Haha. I actually just meant the new specs but it's great that you did them all :-)

I'll take a look at this when I get home.

PR #360, opened 2015-08-05
Feature/jsonapi compound documents

hubert:
all. i've added very basic support for compound documents. 

similar to the earlier commits initial support of JSON API, it moves the included documents into the same namespace as the attributes. however, since JSON API specifies that attributes and included objects occupy the same namespace, there shouldn't be issues with conflicts provided the response is JSON API-compliant.

we're losing some context here by taking this tact, but this allows us to retrieve multiple resources at once in a way that will be compatible with JSON API development efforts down the road. 

the next step will on this front will be to figure out how to generate associations based on the links that are returned rather than relying on a static object that specifies the path. 

feedback very much appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,7 +584,7 @@ users = Users.all
 # GET "/users", response is { "users": [{ "id": 1, "fullname": "Lindsay Fünke" }, { "id": 1, "fullname": "Tobias Fünke" }] }
 ```
 
-#### JSON API support
+#### JSON API support (experimental)
 
 To consume a JSON API 1.0 compliant service, it must return data in accordance with the [JSON API spec](http://jsonapi.org/). The general format
 of the data is as follows: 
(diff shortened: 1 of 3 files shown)


rjayroach:
@hubert First, thanks for taking up as maintainer of Her. much appreciated!

I've been using this branch for some testing with some internal micro services we're building. There are a few features that I want to add to JSON API support including populating an errors hash when the API returns errors, e.g. name cannot be blank. Wondering if you plan on merging this PR in and if you need help what I can do to move it forward?

danielneis:
hello, it would be very usefull.

vaihtovirta:
@hubert Do you have any plans to finish this?
Let me know what's still missing and maybe I can help.

danielneis:
I've tested the changes proposed here and found that they work.
There is a problem with types with "_" or "-", but this is generalized in all her-rb code, so I think it may not be a reason to not get it working. I've experimented with has_one and belongs_to relationships.

Relationships inside other relationships are not parsed.

PR #367, opened 2015-09-10
add reload for associations

zacharywelch:
similar to ActiveRecord [reload](http://apidock.com/rails/ActiveRecord/Associations/Association/reload), this will bust the cache and refetch an association.


zacharywelch:
fixes #375

PR #368, opened 2015-09-10
replace association details w/ association options

zacharywelch:
Replaces any mention of **details** with **options** in associations_spec so it closely matches the description in `association.rb`. nothing glamorous :smile_cat:


zacharywelch:
closing since we have other plans to refactor associations

PR #369, opened 2015-09-22
304 responses don't include message bodies either

ekampp:


hubert:
hi @ekampp. thanks for the contribution!

could you add a spec to cover this behavior? G2G once that's in place.

ekampp:
@hubert Added a spec :+1:

zacharywelch:
Hi @ekampp sorry for the delay. Would you be able to squash these commits before merging? Everything looks good

PR #374, opened 2015-10-01
Feature: Allow payload on destroy actions

### LINKED ISSUE #373: Allow params on destroy
We're sending params over on delete calls, used for soft or hard deletion. At the moment Her does not forward params on `DELETE` actions. I know it's not entirely RESTful. 

This request is mostly up to philosophy, should HTTP DELETE actually contain a payload? Some general discussion can be found here: http://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request

ticktricktrack:
closes #373

### DIFF
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -1,3 +1,4 @@
+# coding: utf-8
 module Her
   module Model
     # This module adds ORM-like capabilities to the model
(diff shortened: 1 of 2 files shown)


hubert:
hi @ticktricktrack. thank you for the contribution.

i don't agree with the practice of passing params to the delete, but i don't see a downside for a general http client such as Her supporting it, since the current HTTP 1.1 RFC sounds agnostic about it. will merge.

PR #378, opened 2015-10-13
Enable customizable JSON serialization on Model and Relation

tcollier:
One use case I have for the Her gem is to retrieve resources from a remote app to be used in the Javascript client of the local app. This entails serializing the Her models into JSON to be returned by a controller in the local app.

This PR introduces `#to_json` and `#as_json` on both the `Model` and `Relation`. `#to_json` will return a nicely formatted JSON string that a Javascript (or other) client of the local app can easily consume.

The interface was explicitly modeled after <a href="http://api.rubyonrails.org/classes/ActiveModel/Serializers/JSON.html">`ActiveModel::Serializers::JSON`</a> and accepts the `:only`, `:except`, and `:methods` options. Keep in mind, this is an incomplete copy of `ActiveModel` functionality, as it does not honor the `:root` or `:include` options. However, it satisfies my current needs and can be extended to accept those options in the future as other developers need them.

While `#to_json` is the primary interface for serializing models and collections, I chose to keep `#as_json` part of the public interface like `ActiveModel` does because it is useful when constructing arbitrary JSON strings, e.g. 

```ruby
response = { user: user.as_json(except: :auth_token), request_id: Request.uuid }
response.to_json
```

### Example usage
```ruby
class User
  include Her::Model

  def best_friend
    # complicated logic to calculate User's best friend
  end
end

user = User.find(1)
p user.to_json(except: [:auth_token])
# {"id":1,name:"Tobias Fünke"}
p user.to_json(methods: "best_friend")
# {"id":1,name:"Tobias Fünke","auth_token":"deadbeef","best_friend":"Lindsay Fünke"}

users = User.all
p users.to_json("only" => "id")
# [{"id":1},{"id":2}]
```

### DIFF
--- a/lib/her/model.rb
+++ b/lib/her/model.rb
@@ -7,6 +7,7 @@ require "her/model/orm"
 require "her/model/parse"
 require "her/model/associations"
 require "her/model/introspection"
+require "her/model/serialization"
 require "her/model/paths"
 require "her/model/nested_attributes"
 require "active_model"
@@ -33,6 +34,7 @@ module Her
     include Her::Model::HTTP
     include Her::Model::Parse
     include Her::Model::Introspection
+    include Her::Model::Serialization
     include Her::Model::Paths
     include Her::Model::Associations
     include Her::Model::NestedAttributes
--- a/lib/her/model/relation.rb
+++ b/lib/her/model/relation.rb
@@ -38,6 +38,22 @@ module Her
       end
       alias all where
 
+      # Return a JSON serialized string representing the fetched collection of models
+      #
+      # @see Model::Serialization#to_json
+      #   for list of allowable options
+      def to_json(options = {})
+        as_json(options).to_json
+      end
+
+      # Map all of the fetched models as Hashes preparing for JSON serialization
+      #
+      # @see Model::Serialization#to_json
+      #   for list of allowable options
+      def as_json(options = {})
+        fetch.map { |model| model.as_json(options) }
+      end
+
       # Bubble all methods to the fetched collection
       #
       # @private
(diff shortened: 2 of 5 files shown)


tcollier:
@hubert It appears you're the maintainer of this gem. I will be building clients for some microservices at work soon and lack of JSON serialization is blocking me from using this gem. Thanks in advance!

hubert:
Hi Tom, Thanks for contributing.

Could you elaborate on the advantages of having Her have it's own serialization routine rather than including ActiveModel::Her::Serialization where needed when json output is required?

Cheers!

tcollier:
Hi Hubert, thanks for responding to my PR.

Including `ActiveModel::Serializers` in a Her model will work for serializing individual models

```ruby
class User
  include Her::Model
  include ActiveModel::Serializers::JSON
end

user = User.find(1)
p user.to_json
# {"id":1,"email":"bob@example.com"}
```

However, it fails when working with relations

```ruby
users = User.all
p users.to_json
# "#<Her::Model::Relation:0x007faec13d46f0>"
```

Let me know if I misunderstood your suggestion.

tcollier:
@hubert (bump)

hubert:
@tcollier thanks for following up and providing an example we can work off of.

i think you understood what i was saying. given the example you show, i think there's a bug. once you call #to_json the relation should resolve and #to_json should be called on the result of actually making the call to retrieve users. 

that said, i don't think explicitly defining as_json and to_json on the relation objects is the right way to go. there does appear to be an incompatibility between Her::Model and ActiveModel::Serialization. if we were able to address that incompatibility, i think your issue would be resolved as well. 

to illustrate, i think if you did the following you'd get the behavior you expect:

```ruby
users = User.all
p users.to_a.to_json
```

tcollier:
@hubert That's a workable solution for me. I'll close this PR, thanks!

PR #381, opened 2015-10-27
Change default behavior of comparison operator like Activerecord

hmatsuda:
her's comparison operator returns false if one object is find from relational methods but Activerecord returns true in same situations.

```
class CustomerResource
  include Her::Model
  has_many :orders
end

class OrderResource
  include Her::Model
  belongs_to :customer
end

> OrderResource.find(1) == OrderResource.find(1)
true
> CustomerResource.orders.find(1) == OrderResource.find(1)
false

class Customer < ActiveRecord::Base
  has_many :orders
end

class Order < ActiveRecord::Base
  belongs_to :customer
end

> Order.find(1) == Order.find(1)
true
> Customer.orders.find(1) == Order.find(1)
true

```

Main cause of that object which includes Her::Model has relation attributes when find from relational resource.

```
Customer.orders.find(1).attributes[:customer_resource]
<CustomerResource(cusotomers/1) price="1000" created_at="2015-10-27T17:01:19.000+09:00" updated_at="2015-10-27T17:02:47.000+09:00" id=1>
```

I think it should be better if her's comparison method works same as Activerecord one.
https://github.com/rails/rails/blob/master/activerecord/lib/active_record/core.rb#L409-L414

### DIFF
--- a/lib/her/model/attributes.rb
+++ b/lib/her/model/attributes.rb
@@ -141,7 +141,7 @@ module Her
       #
       # @private
       def ==(other)
-        other.is_a?(Her::Model) && id == other.id
+        other.is_a?(Her::Model) && self.class == other.class && id == other.id
       end
 
       # Delegate to the == method
--- a/lib/her/model/attributes.rb
+++ b/lib/her/model/attributes.rb
@@ -141,7 +141,7 @@ module Her
       #
       # @private
       def ==(other)
-        other.is_a?(Her::Model) && @attributes == other.attributes
+        other.is_a?(Her::Model) && id == other.id
       end
 
       # Delegate to the == method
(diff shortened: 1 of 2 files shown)


hubert:
thanks for the contribution @hmatsuda.

it doesn't look like the way == is written now is perfect, but the way you've rewritten ==, it would be pretty easy to get false positives.

```
class Foo; include Her::Model; end

class Bar: include Her::Model; end

> Foo.new(id: 1) == Bar.new(id: 1)
true
```

hmatsuda:
@hubert Thanks for reviewing. I added to test both class are same or not into == method and change it's spec.

hubert:
can you add a spec to verify the relation equality from your original example works as expected?

hmatsuda:
@hubert I added a spec. Can you review it?

PR #383, opened 2015-10-29
Implemented attribute_changed_in_place? on models so validates_numericality_of validation will work

glappen:
This seems to work, and specs are passing, but let me know if there's a more natural way to do this.

### DIFF
--- a/lib/her/model/attributes.rb
+++ b/lib/her/model/attributes.rb
@@ -132,6 +132,12 @@ module Her
       end
       alias attribute get_attribute
 
+      # Return `true` if the attribute was changed since last read.
+      # This allows validates_numericality_of to work properly.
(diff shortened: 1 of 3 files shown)


glappen:
For our information, what is the current status of the project?  Is it being maintained?  Can we expect pull requests to be merged at some point?  Thanks!

zacharywelch:
hi @glappen thanks for the PR! We're considering an update to `ActiveModel` which would give us `attribute_changed_in_place?`Would that cover what you need?

chewi:
I've run into this but I'm not sure about this change as I think `attribute_changed_in_place?` is meant to detect changes to mutable objects that would not be visible in the `changes` hash. A newer ActiveModel would be good but I guess that means Rails 5?

PR #389, opened 2015-12-15
JSON API destroy_existing bug

DocX:
Class method `destroy_existing` didn't work with JSON API responses. It was expecting `:attributes` field but JSON API by default response with `204 No Content`  for `DELETE` method.

This patch consolidates class method `destroy_existing` with object method `destroy` to fix the issue and avoid future issues by reducing duplicated code.

### DIFF
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -75,8 +75,9 @@ module Her
       #   # Called via DELETE "/users/1"
       def destroy(params = {})
         method = self.class.method_for(:destroy)
+        path = request_path(params)
         run_callbacks :destroy do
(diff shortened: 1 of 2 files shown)


edtjones:
hi @DocX  I've just got commit access on this repo and plan to get back on track with maintenance over time - will be in touch soon!

zoltankiss:
@DocX can you please let me know if this is still an issue? I tried running your test case against latest master and it was green

DocX:
@zoltankiss Hi. Sadly I don't work with `her` gem anymore, so I can't confirm. If you say the test is green, then it is probably been fixed elsewhere 👍 It may still make sense to add the test to the `master` - looking quickly the `destroy_existing` method is still not being covered?

Issue #401, opened 2016-04-14
no implicit conversion of nil into Hash


avdmatamoros:
TypeError:
       no implicit conversion of nil into Hash
     # /usr/local/rvm/gems/ruby-2.2.3@bodhi/gems/her-0.8.1/lib/her/model/parse.rb:36:in `merge!'
     # /usr/local/rvm/gems/ruby-2.2.3@bodhi/gems/her-0.8.1/lib/her/model/parse.rb:36:in `to_params'
     # /usr/local/rvm/gems/ruby-2.2.3@bodhi/gems/her-0.8.1/lib/her/model/parse.rb:13:in `to_params'
     # /usr/local/rvm/gems/ruby-2.2.3@bodhi/gems/her-0.8.1/lib/her/model/orm.rb:43:in `block (2 levels) in save'

Avoid using next inside inject blocks, this can yield a nil instead of a hash, instead treat it as a normal if statement.

https://github.com/remiprev/her/blob/master/lib/her/model/parse.rb#L61

PR #402, opened 2016-04-14
Fix object serialization

dmatamoroscr:
Prevent inject from returning nil

https://github.com/remiprev/her/issues/401

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 /pkg
 /tmp
 /coverage
+
+.idea
--- a/lib/her/api.rb
+++ b/lib/her/api.rb
@@ -97,7 +97,8 @@ module Her
         else
           # For POST, PUT and DELETE requests, treat additional parameters as request body
           request.url path
-          request.body = opts
+          # Get json representation object and convert to json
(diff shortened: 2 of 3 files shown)


mtwentyman:
Another option to help avoid an incorrect return type (while allowing the use of next) is to use `Enumerable#each_with_object`.  It might have a slight side benefit of making it a little harder for this to break in the future.  Speaking of which, I'd love to see a spec for this too.  From the [contribution guide](https://github.com/remiprev/her/blob/master/CONTRIBUTING.md):

> Add examples that describe it (in the spec directory). Only refactoring and documentation changes require no new tests. If you are adding functionality or fixing a bug, we need examples!

cimtico:
I'm not entirely sure how to build a spec for This scenario. Also it might even be overkill considering it's a bad usage of inject.
Regarding the each with object, I agree it might be a better solution to the inject in this case.

edtjones:
hi @avdmatamoros @mtwentyman @cimtico I've just got commit access on this repo and plan to get back on track with maintenance over time - will be in touch soon!

PR #412, opened 2016-08-12
Return an empty collection when where(:foo => []) is given

chewi:
The previous behaviour was to return everything! This is now consistent with ActiveRecord and avoids a needless request.

The new behaviour is not applied to has_one associations as they currently seem to ignore the where method entirely.


edtjones:
hi @chewi I've just got commit access on this repo and plan to get back on track with maintenance over time - will be in touch soon!

chewi:
Rebased.

chewi:
Rebased again.

chewi:
And again!

PR #418, opened 2016-10-25
JsonApi relationships data accessible from attributes now.

mehmetdavut:
relationships data was not available, so with Her::JsonApi::Model is
not possible to create relations between models.


edtjones:
Thanks for this @mehmetdavut - we're still getting up to speed with reviewing all the PRs but will get to this one asap.

boena:
Any reason this isn't merged? Seems like a vital point to be able to access relations when using JSON:API

PR #423, opened 2016-12-02
Fixes embedded_params when params are empty

masonjm:
`params` expects `embeded_params` to always return a `Hash`. This fixes an edge case where `embeded_params` can return `nil`, which causes an exception in `merge!`.

### DIFF
--- a/lib/her/model/parse.rb
+++ b/lib/her/model/parse.rb
@@ -58,7 +58,7 @@ module Her
         def embeded_params(attributes)
           associations[:has_many].select { |a| attributes.include?(a[:data_key])}.compact.inject({}) do |hash, association|
             params = attributes[association[:data_key]].map(&:to_params)
-            next if params.empty?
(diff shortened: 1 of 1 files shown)


edtjones:
@masonjm thanks for this. How did you find this edge case? Would it be best to fix the reason there isn't a hash in the first place?

masonjm:
@edtjones we discovered this when upgrading from 0.7.6 to 0.8. We had a bunch of specs start failing with obscure errors. We eventually tracked it down to a problem with some seed data in our specs.

It's possible this fix isn't necessary, but tracking down the source of the error this generates was non-trivial. It also bothered me that the code could fail in an expected way, and nothing was checking for that :).

edtjones:
@masonjm so the `next` was producing nils whereas your `next hash` fix returns a hash. Nice. Seems like it shouldn't break anything. I'm happy if you're happy :-)

PR #424, opened 2016-12-02
Update associated record creation example

mascott:
While the example is probably the intended behavior it does not reflect the current request path. I've updated the example to more closely match the current experience.

If one of the potential fixes like https://github.com/remiprev/her/pull/318 is found to be acceptable it would make sense to include an update to the READ.me to reflect the changes.


edtjones:
Thanks @mascott I'm grateful. Merged :-)

PR #428, opened 2017-01-10
Handle the destroyed instance variable according to errors

dhenze:
The presence of response_errors for a destroy request indicates the
request has not been successful, e.g. due to authorisation issues. By
returning a truthy value a false indication is given as we cannot
assume everyone checking for response errors.

I therefore relate the value of ```@destroyed``` to the existence of errors
as it seems more sensible to me.

Looking forward to your comments. Especially since this may be a
breaking change for some users.

### DIFF
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -80,7 +80,7 @@ module Her
             assign_attributes(self.class.parse(parsed_data[:data])) if parsed_data[:data].any?
             @metadata = parsed_data[:metadata]
             @response_errors = parsed_data[:errors]
-            @destroyed = true
+            @destroyed = @response_errors.empty?
           end
         end
         self
@@ -163,7 +163,7 @@ module Her
             data = parse(parsed_data[:data])
             metadata = parsed_data[:metadata]
             response_errors = parsed_data[:errors]
-            new(data.merge(:_destroyed => true, :metadata => metadata, :response_errors => response_errors))
+            new(data.merge(:_destroyed => response_errors.empty?, :metadata => metadata, :response_errors => response_errors))
           end
         end
 
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -80,7 +80,7 @@ module Her
             assign_attributes(self.class.parse(parsed_data[:data])) if parsed_data[:data].any?
             @metadata = parsed_data[:metadata]
             @response_errors = parsed_data[:errors]
-            @destroyed = @response_errors.empty?
+            @destroyed = response.success?
           end
         end
         self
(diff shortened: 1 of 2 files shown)


edtjones:
Thanks for this @dhenze - I think if we're going to return errors for the destroy method shouldn't we do so for all request types?

dhenze:
@edtjones I actually did not change the behaviour of the response_errors, instead I do set the ```@destroyed``` value of the resource accordingly. My assumption here is: An error occured during the delete request, the resource therefore has not been deleted and the resource should not behave like it has been deleted by returning true for ```.destroyed?```
To be honest, I haven't checked for other request types, but most likely similar behaviour may occur for POST (e.g. ```persisted?``` returning a truthy value even though creation failed with errors) and PUT requests. But as stated, haven't checked.
First and foremost I would like to clarify this is the way we want to proceed.

dhenze:
I did update the PR to refer to the response status rather then requiring the errors to be set. I guess this is a much more generic approach.

edtjones:
@ticktricktrack @zkwentz @foxpaul do you have a view? Seems smart to have responses which represent what happened, UNLESS we want to take the approach that the thing returning the success was the _request_ in which case what happens now is correct.

dhenze:
Hi there, any news on the discussion?

edtjones:
Sorry @dhenze haven't had headspace to review this. Looks ok from a quick look. Please nudge if you haven't heard anything in a week.

dhenze:
@edtjones None of the others seems to have considered this PR. Should I outline what this PR is about more clearly?

edtjones:
Hi @dhenze I think we should merge this. I think the general expectation would be to return errors from the api. Will do it tonight.

dhenze:
Thanks @edtjones Very much appreciated. And yes, signaling a success without checking the response causes false expectations. That could lead to other transactional actions in a service ending in inconsistencies.

dhenze:
@edtjones Any news on this?

edtjones:
So sorry - had a bananas time the last few weeks. Will merge asap. 

Would you like commit access? I am keen to share the work with people who want to be involved. 

Ed

--
Sent from my phone: please forgive the brevity.

Ed Jones

edtjones:
@dhenze with big apologies for the delay, I've merged this and released 0.8.5 to rubygems. Let me know if you want commit access - I'm keen to share the work :-)

dhenze:
@edtjones Thanks a lot! 
As for the commit access, as soon as I see something popping up and the need to return, I will get in  contact with you to set something up. Right now it looks like a one-off effort.

dhenze:
Btw. the fix originates of work from @lipanski. 🙇

Issue #252, opened 2014-06-08
Best way to authenticate with HTTP Basic?


alainmeier:
Our authentication system works like Stripe's wherein the username field in HTTP Basic Auth has to be the user's API key. So I am trying to set this up with Her, but I can't seem to get it quite right. I created some middleware:

``` ruby
class ApiKeyAuthentication < Faraday::Middleware
  def call(env)
    username = Base64.encode64(RequestStore.store[:user_api_key])
    env[:request_headers]["Authorization"] = "Basic " + Base64.encode64(username + ":")
    @app.call(env)
  end
end
```

And I set the `RequestStore.store[:user_api_key]` in the ApplicationController. Then I included the middleware in the her config file. I cannot set the API key directly from the config file (like #50) because it has to be supplied by the ApplicationController. However, it appears that this does not authenticate properly. What is the best way to authenticate with HTTP Basic?

zacharywelch:
README updated in #431

Issue #50, opened 2012-11-20
Problems with the example


ceicke:
Hi,

I am trying to use the 'her' gem as it seems like a really good gem. However I can't even get the simplest thing to work. Here is the error I am getting

``` ruby
1.9.2p320 :001 > s = Fileflow::Student.find(1)
URI::InvalidURIError: query conflicts with opaque
    from /Users/ceicke/.rvm/rubies/ruby-1.9.2-p320/lib/ruby/1.9.1/uri/generic.rb:521:in `check_query'
    from /Users/ceicke/.rvm/rubies/ruby-1.9.2-p320/lib/ruby/1.9.1/uri/generic.rb:540:in `query='
    from /Users/ceicke/.rvm/gems/ruby-1.9.2-p320@intranet/gems/faraday-0.8.4/lib/faraday/connection.rb:277:in `build_exclusive_url'
    from /Users/ceicke/.rvm/gems/ruby-1.9.2-p320@intranet/gems/faraday-0.8.4/lib/faraday/request.rb:94:in `to_env'
    from /Users/ceicke/.rvm/gems/ruby-1.9.2-p320@intranet/gems/faraday-0.8.4/lib/faraday/connection.rb:225:in `run_request'
    from /Users/ceicke/.rvm/gems/ruby-1.9.2-p320@intranet/gems/faraday-0.8.4/lib/faraday/connection.rb:87:in `get'
    from /Users/ceicke/.rvm/gems/ruby-1.9.2-p320@intranet/gems/her-0.2/lib/her/api.rb:67:in `request'
    from /Users/ceicke/.rvm/gems/ruby-1.9.2-p320@intranet/gems/her-0.2/lib/her/model/http.rb:37:in `request'
    from /Users/ceicke/.rvm/gems/ruby-1.9.2-p320@intranet/gems/her-0.2/lib/her/model/orm.rb:53:in `find'
    from (irb):1
    from /Users/ceicke/.rvm/gems/ruby-1.9.2-p320@intranet/gems/railties-3.0.10/lib/rails/commands/console.rb:44:in `start'
    from /Users/ceicke/.rvm/gems/ruby-1.9.2-p320@intranet/gems/railties-3.0.10/lib/rails/commands/console.rb:8:in `start'
    from /Users/ceicke/.rvm/gems/ruby-1.9.2-p320@intranet/gems/railties-3.0.10/lib/rails/commands.rb:23:in `<top (required)>'
    from script/rails:6:in `require'
[...]

PR #431, opened 2017-01-20
Add notes about basic http auth

jhirbour:
I was setting this Gem up and needed to use http auth. 

I found references in https://github.com/remiprev/her/issues/50 and https://github.com/remiprev/her/issues/252 . I figured it makes sense to add it to the Readme.


edtjones:
Thanks @jhirbour - I'm grateful :-)

PR #433, opened 2017-02-01
Added support of unchanged requests

kvokka:
Added middleware for correct support status 304 in responses

It gives huge speed boost, if you have to

request same resources offten


edtjones:
Thanks @kvokka - any reason for writing this rather than using the https://github.com/plataformatec/faraday-http-cache gem?

Also, I wonder if you can fix up the failing tests?

kvokka:
Thank you for so quick response.

gem https://github.com/plataformatec/faraday-http-cache is great, but it targets to client-side optimizing, rather this middleware is targeting to back-end.

This chink was made to prevent heavy response generation on other side, in favor checking `updated_at` field only, so, 200-500ms sub-requests transformed to 2-5ms.

Travis is green now :)

I've fix syntax to the most common style for Ruby 1.9.3 support. Strange to see 1.9.3 in 2017

kvokka:
In Caching part of the README shown way, how to implement caching, but, if on other side object will be changed there is no way to get this information. Only `expire_in` cache, so you will do lots of unneeded expiration's or use stale data. With this middleware there is the way on other side of back-end microservice to know, since what moment the object is fresh.

Of course it is not perfect solution, but it is much more livable, than that, which is included in README.

If you like the code, I can do README update also (or you may take comments from the middleware)

edtjones:
@kvokka a quick holding message to say I haven't forgotten about this! Just want to find some time to review properly.

edtjones:
@zacharywelch if you're feeling keen could you review this?

zacharywelch:
Hi @kvokka, thanks for the PR! Do you mind putting together some specs describing the middleware  behavior? See [first_level_parse_json_spec](https://github.com/remiprev/her/blob/master/spec/middleware/first_level_parse_json_spec.rb) for an example.

I'm also curious what the difference would be with something like faraday-http-cache where it would cache api responses and check whether they were stale on subsequent requests to another microservice or api. Do you have an example you could share?

[Another idea](https://github.com/remiprev/her/pull/445#issuecomment-305936181) discussed in #445 would be to package your own faraday middleware if it's something that could be used by other faraday consumers. Generally **her** middleware is fundamental to the ORM.

PR #436, opened 2017-03-08
Allow to parse response without attributes key

pablocrivella:


edtjones:
HI @pablocrivella thanks for the PR! What situation are you trying to resolve here? Why wouldn't you have an attributes key?

pablocrivella:
I have some endpoints that only return id and type, and according to JSON API spec attributes is not a required field.

pablocrivella:
Hey @edtjones any update about this? :)

PR #439, opened 2017-05-19
Support the latest Ruby version on Travis CI

yensaki:
Travis CI against the latest Ruby 2.4.1.
I added it.


edtjones:
Thanks @yensaki that's brilliant. I've merged. Won't deploy a new gem version for now. (Unless you think I should?)

PR #440, opened 2017-05-29
add reload to model

zacharywelch:
Similar to [ActiveRecord::Persistence.reload](https://apidock.com/rails/ActiveRecord/Persistence/reload), this will refetch a resource from the API and assign its attributes in-place.

```ruby
user = User.find(1)
# => #<User(users/1) id=1 name="Tobias Fünke">
user.name = "Oops"
user.reload # Fetched again via GET "/users/1"
# => #<User(users/1) id=1 name="Tobias Fünke">
```
As described in the docs
> Reloading is commonly used in test suites to test something is actually
> written to the database, or when some action modifies the corresponding
> row in the database but not the object in memory

It's useful in the console as well.

fyi PR #367 also adds `reload` to associations 😺

### DIFF
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -86,6 +86,23 @@ module Her
         self
       end
 
+      # Refetches the resource
+      #
+      # This method finds the resource by its primary key (which could be
+      # assigned manually) and modifies the object in-place.
+      #
+      # @example
+      #   user = User.find(1)
+      #   # => #<User(users/1) id=1 name="Tobias Fünke">
+      #   user.name = "Oops"
+      #   user.reload # Fetched again via GET "/users/1"
+      #   # => #<User(users/1) id=1 name="Tobias Fünke">
+      def reload(options = nil)
+        fresh_object = self.class.find(id)
+        assign_attributes(fresh_object.attributes)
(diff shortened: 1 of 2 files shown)


edtjones:
This is brilliant @zacharywelch thank you! I'm quite snowed under; how confident should I be just merging this? Will it break anything? I can't see that it can really.

edtjones:
(From a quick code review it looks really self-contained)

zacharywelch:
Hi @edtjones you're right it's rather self-contained and shouldn't interfere w/ her's public api. I spent some time confirming the `Relation` params weren't carried over after `reload` and that looks fine as well 👍

edtjones:
Amazing. I will try to merge and release this week. Thanks again!

(Would you like commit access? I'm keen to share with as many as are keen to contribute)

zacharywelch:
sounds awesome! sent an email to your `errorstudio` address 😄

vinc:
Thanks for this PR! Do you think you can also merge #367?

I started using Her last week and I was just looking for a way to reload assotiations in my test suit.

edtjones:
@vinc yes! Plan to merge @zacarywelch's stuff asap.

PR #442, opened 2017-05-29
add toggle and toggle! to model

zacharywelch:
Similar to [ActiveRecord::Persistence.toggle](https://apidock.com/rails/ActiveRecord/Persistence/toggle), this will assign to `attribute` the boolean opposite of `attribute?` without calling save. There's also a `toggle!` version which saves the resource afterwards.

```ruby
user = User.first # => #<User(users/1) id=1 admin=false>
user.admin? # => false
user.toggle(:admin)
user.admin? # => true

user.reload.admin? # => false
user.toggle!(:admin) # => true
# Saved via PUT "/users/1"
user.admin? # => true
```

### DIFF
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -103,6 +103,27 @@ module Her
         self
       end
 
+      # Assigns to +attribute+ the boolean opposite of <tt>attribute?</tt>. So
+      # if the predicate returns +true+ the attribute will become +false+. This
+      # method toggles directly the underlying value without calling any setter.
+      # Returns +self+.
+      #
+      # @example
+      #   user = User.first
+      #   user.admin? # => false
+      #   user.toggle(:admin)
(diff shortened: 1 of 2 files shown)


edtjones:
@zacharywelch can you sort out the conflict in orm.rb before I merge this? Also an update to the Readme would be ace. Thanks :-)

zacharywelch:
Rebased 👍 

Didn't see an obvious section in the readme for `toggle` but then too many examples would probably pollute the document? However we do have information in the [rubydocs](http://www.rubydoc.info/github/remiprev/her/Her/Model/ORM#reload-instance_method) for `reload`,
 `toggle`, `first_or_create` etc. 😺

edtjones:
Ace - you want to merge these? You have commit :-). I probably can't release a new gem til Tuesday I'm afraid.

PR #400, opened 2016-04-07
Adds Faraday options "timeout" and "open_timeout" accessible in her config

abhikalakuntla:
Contrary to the the comments in this issue: https://github.com/remiprev/her/issues/120, the options `timeout` and `open_timeout` are not being accepted by Her.

Github page for [Faraday](https://github.com/lostisland/faraday)

PR #449, opened 2017-06-07
Fix for Faraday.new options

stereobooster:
It expexts 1..2 parameters: def new(url = nil, options = nil)

### DIFF
--- a/spec/api_spec.rb
+++ b/spec/api_spec.rb
@@ -21,10 +21,10 @@ describe Her::API do
       end
 
       context "when setting custom options" do
-        before { subject.setup foo: { bar: "baz" }, url: "https://api.example.com" }
+        before { subject.setup foo: { bar: "baz" }, url: "https://api.example.com", timeout: 10, open_timeout: 2 }
 
         describe "#options" do
-          it { expect(subject.options).to eq(foo: { bar: "baz" }, url: "https://api.example.com") }
+          it { expect(subject.options).to eq(foo: { bar: "baz" }, url: "https://api.example.com", timeout: 10, open_timeout: 2) }
         end
       end
     end
--- a/gemfiles/Gemfile.activemodel-3.2.x
+++ b/gemfiles/Gemfile.activemodel-3.2.x
@@ -4,4 +4,4 @@ gemspec :path => "../"
 
 gem 'activemodel', '~> 3.2.0'
 gem 'activesupport', '~> 3.2.0'
-gem 'faraday', '~> 0.8.9'
+gem 'faraday'
--- a/gemfiles/Gemfile.activemodel-4.0
+++ b/gemfiles/Gemfile.activemodel-4.0
@@ -4,4 +4,4 @@ gemspec :path => "../"
 
 gem 'activemodel', '~> 4.0.0'
 gem 'activesupport', '~> 4.0.0'
-gem 'faraday', '~> 0.8.9'
+gem 'faraday'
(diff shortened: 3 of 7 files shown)


edtjones:
@zacharywelch you ok with this?

spec/api_spec.rb
```diff
@@ -21,10 +21,10 @@ class Bar; end
       end
 
       context "when setting custom options" do
-        before { subject.setup foo: { bar: "baz" }, url: "https://api.example.com" }
+        before { subject.setup foo: { bar: "baz" }, url: "https://api.example.com", timeout: 10, open_timeout: 2 }
```
stereobooster: this test shows error - without fix it will fail

zacharywelch: You should be seeing a green build. The travis build passes on your [first commit](https://travis-ci.org/remiprev/her/builds/240389166?utm_source=github_status&utm_medium=notification) 😃

zacharywelch: So the build is passing on Travis w/ your [first commit](https://travis-ci.org/remiprev/her/builds/240388060) and  to be sure I tested on a separate [branch](https://travis-ci.org/remiprev/her/builds/240842291). We should be ok  to drop commit 6f01adce19caa36c24b4a7ec931e99aa6fc4578d 👍

stereobooster: Here is failing test case https://travis-ci.org/remiprev/her/jobs/241677947

stereobooster:
Test didn't fail for Travis, but failed for local env because Gemfiles used by two envs are different. We can change Gemfile to:

```ruby
eval_gemfile(File.join(File.dirname(__FILE__), "gemfiles/Gemfile.activemodel-4.2"))
```

This way we will make sure that it is at least someway intersect.

spec/api_spec.rb
zacharywelch: Believe there's confusion around the `timeout` and `open_timeout` options introduced in #400. These options should be part of the [request hash](https://github.com/lostisland/faraday/blob/a712938071bb0997410b5de4a4f2517d30fc4d3d/lib/faraday/options.rb#L192) set on the connection. There's a [thread](https://github.com/lostisland/faraday/issues/417#issuecomment-223413386) on Faraday w/ more detail.

As far as I can tell `open_timeout` and `timeout` are never recognized by Faraday unless they're part of the request hash, going as far back as `0.5.7`. The syntax should be

```ruby
Her::API.setup(url: "https://api.example.com", request: { timeout: 10, open_timeout: 2 })
```

Can you confirm?

stereobooster: Sorry, still haven't tested it. But found another connected bug: [Net::HTTP retries idempotent requests once after a timeout, but its not configurable](https://bugs.ruby-lang.org/issues/10674). So if we will fix this bug, and I will pass timeout, Net::HTTP will do two requests in case of Timeout. Nice.

zacharywelch: Haven't played around w/ the faraday [retry options](https://github.com/lostisland/faraday/blob/master/lib/faraday/request/retry.rb#L25) much. Is there anything there you could take advantage of? 

Regardless it sounds like this is a separate issue and there's a separate [PR](https://github.com/remiprev/her/pull/452) explaining how to use `timeout` and `open_timeout`. I'll go ahead and close this PR then.

zacharywelch:
Going to close this for now but feel free to reopen if the suggestions in #452 are insufficient. Thanks!

PR #450, opened 2017-06-07
Fix for correct ActiveModel::Dirty #previous_changes implementation

gotmayonase:
ActiveModel::Dirty specifies that `#previous_changes` should respond with the same output as `#changes` did prior to the save, including both the value of the field before and after the save.  The implementation in Her was responding with `#changed_attributes`, which only gives the original value of the field, not before and after changing.

This should resolve that discrepancy.

### DIFF
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -49,7 +49,7 @@ module Her
 
               return false if !response.success? || @response_errors.any?
               if self.changed_attributes.present?
-                @previously_changed = self.changed_attributes.clone
+                @previously_changed = self.changes.clone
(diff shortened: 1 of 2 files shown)


zacharywelch:
@gotmayonase mind squashing these commits real quick before merging?

zacharywelch:
Thanks!

edtjones:
@gotmayonase @zacharywelch thanks for this! 👍

gotmayonase:
@edtjones Happy to help! 😄

gotmayonase:
Also thanks for that amazingly fast response time @zacharywelch!

PR #454, opened 2017-06-12
refactor of attributes

zacharywelch:
`attributes.rb` needs a little routine maintenance and tlc 💅 💅 

- reuse regular expression for predicate and writer methods in `respond_to_missing`
- move `initialize_collection` and `use_setter_methods` to ClassMethods section with other class methods
- rename `initialize_collection` to `instantiate_collection` and add an `instantiate_record` method similar to [ActiveResource](https://github.com/rails/activeresource/blob/c38f18ce4ee6c83e034e7b465459fdd547804b64/lib/active_resource/base.rb#L1061).
- replace duplication of `new_from_parsed_data` and `instantiate_record` from `http` and other files
- apply 80 character limit for readability

### DIFF
--- a/lib/her/model/attributes.rb
+++ b/lib/her/model/attributes.rb
@@ -139,20 +139,29 @@ module Her
 
       module ClassMethods
 
+        # Initialize a single resource
+        #
+        # @private
+        def instantiate_record(klass, parsed_data)
+          if record = parsed_data[:data] and record.kind_of?(klass)
+            record
+          else
+            attributes = klass.parse(record).merge(_metadata: parsed_data[:metadata],
+                                                   _errors: parsed_data[:errors])
+            klass.new(attributes).tap do |record|
+              record.run_callbacks :find
+            end
+          end
+        end
+
         # Initialize a collection of resources
         #
         # @private
-        def instantiate_collection(klass, parsed_data={})
-          collection_data = klass.extract_array(parsed_data).map do |item_data|
-            if item_data.kind_of?(klass)
-              resource = item_data
-            else
-              resource = klass.new(klass.parse(item_data))
-              resource.run_callbacks :find
-            end
-            resource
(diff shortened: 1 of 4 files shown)


shashiprab:
Any movement on this? The longer it sits around, the more stale it gets.

zacharywelch:
Hi, was trying to defer these changes until we had a chance to release 0.9.0. Now that's taken care of perhaps we'll merge this weekend. Until then feel free to add any comments or give it a review 👍

shashiprab:
Cool. Appreciate your efforts that have moved the needle over the last month tremendously! 

-Shashi

shashiprab:
Zachary

If you could share your email with me I have proposal about the Her gem. My email is shaship@yahoo.com 

Thanks

-Shashi

lib/her/model/attributes.rb
```diff
@@ -177,42 +144,92 @@ def attribute?(attribute)
...
+        end
+
+        # Use setter methods of model for each key / value pair in params
+        # Return key / value pairs for which no setter method was defined on the
+        # model
+        #
+        # @private
+        def use_setter_methods(model, params = {})
+          reserved = [:id, model.class.primary_key, *model.class.association_keys]
+          model.class.attributes *params.keys.reject { |k| reserved.include?(k) }
```
huangjun604: Hi @zacharywelch , is there any reason to ignore the string keys?

zacharywelch: Good catch 👍Reserved keys are stored as symbols but attributes can be assigned with string keys. It should instead read 

```ruby
reserved.include?(k.to_sym)
```

Would you like to submit a PR?

huangjun604: Thanks for response @zacharywelch . Please help to update it, thanks.

Issue #422, opened 2016-11-28
Her requests has_many association when parent ID is blank.


jkappers:
Is it intended for Her to make this association GET request even though `pizza.id.blank?`?

```ruby
class Pizza
  include Her::Model
  has_many :fixins
end

class Fixin
  include Her::Model
  belongs_to :pizza
end

##
# Routes available:
# GET /pizzas
# GET /pizzas/:id/fixins
#
p = Pizza.new
p.fixins # GET /pizzas/fixins 404
```

edtjones:
Hi @jkappers yeah that feels wrong doesn't it. Will mark as a bug.

edtjones:
@jkappers in practical terms what problem is this causing you? And what would you have expected?

jkappers:
@edtjones Thanks for getting back to me so quickly!

It's nothing that I can't work around. I discovered it by doing something similar to: 

```ruby
class User
  has_many :phone_numbers
 
  validate :at_least_one_phone_number

  def at_least_one_phone_number
    if self.phone_numbers.count < 0 || self.phone_numbers.none?(&:valid?)
      # errors.add or whatever.
    end
  end
end

class PhoneNumber
  belongs_to :user
end

u = User.new
u.valid? # NoMethodError: undefined method `count' for #<PhoneNumber(phone_numbers) >
```

My expectation was something like:

```ruby
def phone_numbers
  return @cached_result if @cached_result
  return Her::Collection.new unless parent.persisted?
  @klass.get
end
```

I wasn't sure if the behavior I was experiencing was by design or not.

edtjones:
@jkappers thanks. The relations aren't an area of the code I know well, but I'll do a bit of spelunking when I can. @zkwentz @ticktricktrack d'you recognise this issue?

zacharywelch:
Nice catch @jkappers. Let us know if you want to create a PR for this behavior :smiley:

zacharywelch:
Fixed in #456. Thanks!

PR #456, opened 2017-06-15
Don't attempt to fetch an association's data if the parent is a new record

gotmayonase:
Here's my attempt to solve issue https://github.com/remiprev/her/issues/422.

Associations are a complicated beast, so it's quite possible I'm missing some part of the picture.  But the tests pass ;)


gotmayonase:
@zacharywelch My pleasure 😄

PR #460, opened 2017-09-11
allow model to have attribute that is the same as its name

att14:
We have a model called `Rate` that has an attribute legacy `rate` which is causing issues when using this gem. This change checks the type of the key to make sure it should be parsed as an object or treated as an attribute.

I also added `pry` because it is useful for debugging. Let me know if you want me to remove it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 /pkg
 /tmp
 /coverage
+.bundle
+/vendor
--- a/.pryrc
+++ b/.pryrc
@@ -1,0 +1,4 @@
+Pry.commands.alias_command 'c', 'continue'
+Pry.commands.alias_command 's', 'step'
+Pry.commands.alias_command 'n', 'next'
+Pry.commands.alias_command 'f', 'finish'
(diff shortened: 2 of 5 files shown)


edtjones:
Thanks @att14 - @zacharywelch are we happy to require pry as a development dep?

@att14 doesn't your change mean that top level attributes might be ignored? (Sorry - doing a quick review on my phone so I might have misunderstood)

att14:
No, this should not ignore keys. This function is used when determining if the object is nested. E.g.

```json
{
  "rate": {
    "rate": "$1.23"
  }
}
```

Vs

```json
{
  "rate": "$1.23"
}
```

Currently in the later case this function believes the object is nested because the `root_element` does exist but it is not an object. The code I've added just checks that it is an object or list of objects.

This isn't a perfect change and I would suggest that long term this become explicitly defined by the user like the other class methods mixed in by `Parse`.

zacharywelch:
Would rather leave `pry` out, though others are free to use it locally 😃 

Believe what @att14 is proposing is allowing attributes with the same name as the possible root element? Sounds reasonable 👍 

##### without root element
```http
GET /roles/1 
{ "role": "admin" }
```
##### with root element
```http
GET /roles/1 
{ "role": { "role": "admin" } }
```

edtjones:
@att14 would you mind rebasing and removing the pry dependency? Happy to merge after that.

zacharywelch:
@att14: went ahead and amended the commit before merging. Thanks again for your contributions on this!

PR #470, opened 2017-11-03
Add support for the OPTIONS verb

edtjones:
Add support for the OPTIONS verb, which requires us to use Faraday's run_request() method instead of the dynamically-generated method associated with the verb name (Faraday.options is a different thing - see lostisland/faraday#305)

@zacharywelch are you ok for me to merge into master and push a 0.9.1 gem (need it for a project we're working on)?

### DIFF
--- a/lib/her/api.rb
+++ b/lib/her/api.rb
@@ -89,19 +89,26 @@ module Her
       path = opts.delete(:_path)
       headers = opts.delete(:_headers)
       opts.delete_if { |key, value| key.to_s =~ /^_/ } # Remove all internal parameters
-      response = @connection.send method do |request|
+      # Faraday doesn't support the OPTIONS verb because of a name collision with an internal options method
+      # so we need to call run_request directly.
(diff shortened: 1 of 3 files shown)


lib/her/api.rb
```diff
@@ -89,19 +89,26 @@ def request(opts={})
       path = opts.delete(:_path)
       headers = opts.delete(:_headers)
       opts.delete_if { |key, value| key.to_s =~ /^_/ } # Remove all internal parameters
-      response = @connection.send method do |request|
+      # Faraday doesn't support the OPTIONS verb because of a name collision with an internal options method
```
zacharywelch: do we want to put these comments underneath `if method == :options` like we do for `if method == :get`?

edtjones:
@zacharywelch I decided to bump to version 0.10.0 rather than 0.9.1 because this release includes your refactor of attributes which is a relatively big chunk of stuff. Hope ok.

PR #472, opened 2017-11-22
Update the setter methods cache after defining new attributes

chewi:
Not doing so resulted in very strange behaviour. Any previously unseen attributes passed to `new` with `send_only_modified_attributes` enabled would be missing from the request parameters. This is because change tracking is only effective for setter methods.

Now this is fixed, it is unclear to me what the `@attributes.merge!` line in `assign_attributes` is for. In theory, all the given attributes (that are not associations) should have setter methods defined by this point. It should probably still trigger change tracking in any case.

### DIFF
--- a/lib/her/model/attributes.rb
+++ b/lib/her/model/attributes.rb
@@ -239,7 +239,11 @@ module Her
         #   end
         def attributes(*attributes)
           attribute_methods_mutex.synchronize do
-            define_attribute_methods attributes
+            old_methods = generated_attribute_methods.instance_methods
+            define_attribute_methods(attributes).tap do
+              new_methods = generated_attribute_methods.instance_methods
+              @_her_setter_method_names = nil if old_methods != new_methods
+            end
           end
         end
 
--- a/spec/model/attributes_spec.rb
+++ b/spec/model/attributes_spec.rb
@@ -393,4 +393,21 @@ describe Her::Model::Attributes do
       end
     end
   end
+
+  it 'updates the cache of setter methods' do
+    spawn_model "Foo::User" do
+      attr_reader :setter_called
+    end
+
+    Foo::User.new(fullname: "Tobias Fünke")
+
+    Foo::User.class_eval do
+      def password=(value)
+        @setter_called = true
+      end
+    end
+
+    user = Foo::User.new(password: 'supersecret')
+    expect(user.setter_called).to be_truthy
+  end
 end


chewi:
Having thought about this some more, I'm wondering why we're caching these at all. I expected to find that `respond_to?` is relatively fast and that any small benefit this might have had would be lost following this change. This is a somewhat unscientific test but my results show that `respond_to?` is faster, even without this change.

```ruby
require 'benchmark'
require 'set'

times = 10_000_000
array = [:foo, :bar]
method = :first
methods = array.methods.to_set

Benchmark.bm(20) do |bm|
  bm.report('Set#include?') { times.times { methods.include?(method) } }
  bm.report('Object#respond_to?') { times.times { array.respond_to?(method) } }
end
```

```
                           user     system      total        real
Set#include?           0.980000   0.000000   0.980000 (  0.980424)
Object#respond_to?     0.810000   0.000000   0.810000 (  0.810658)
```

chewi:
Aha. That was with Ruby 2.4. With 2.3, the results are reversed but there's still not much in it.

```
                           user     system      total        real
Set#include?           0.890000   0.000000   0.890000 (  0.892254)
Object#respond_to?     1.030000   0.000000   1.030000 (  1.036518)
```

With JRuby 9.1.13.0, `responds_to?` wins by a bigger margin.

```
                           user     system      total        real
Set#include?           1.620000   0.000000   1.620000 (  1.102858)
Object#respond_to?     0.690000   0.000000   0.690000 (  0.562755)
```

I'm going to open an alternative pull request to remove the cache entirely.

chewi:
Okay, I now see why `respond_to?` isn't used. Because of `respond_to_missing?`, it returns true for any method ending in `=`. Unfortunately calling `methods` on every iteration is proving very expensive and that explains why the cache was added. I guess this needs a rethink...

chewi:
I have two ideas.

1. Use a fibre-local variable to make `respond_to_missing?` return false while doing these checks. This feels a little ugly but otherwise has no downsides.
2. Check whether `model.method(:foo=).source_location` returns nil or raises a `NoMethodError`. The problem with this is that `source_location` also returns nil for native code.

I much prefer the first one. They're both slower than the earlier approaches but much faster than calling `methods` repeatedly. The first one gets a small boost if you reset the variable to false rather than nil, presumably because nil deletes the variable, requiring it to be recreated next time. On a side note, passing a string to `respond_to?` is significantly slower than passing a symbol.

```ruby
require 'benchmark'
require 'set'

class Foo
  def bar=(value)
  end

  def respond_to_missing?(method, include_private = false)
    !!Thread.current[:her_respond_to_missing]
  end

  def respond_to_without_missing_nil?(method, include_private = false)
    Thread.current[:her_respond_to_missing] = true
    respond_to?(method, include_private)
  ensure
    Thread.current[:her_respond_to_missing] = nil
  end

  def respond_to_without_missing_false?(method, include_private = false)
    Thread.current[:her_respond_to_missing] = true
    respond_to?(method, include_private)
  ensure
    Thread.current[:her_respond_to_missing] = false
  end
end

times = 10_000_000
foo = Foo.new
method = 'bar='

Benchmark.bm(33) do |bm|
  bm.report('Set#include?') { methods = Foo.instance_methods.map(&:to_s).to_set; times.times { methods.include?(method) } }
  bm.report('Object#respond_to?') { times.times { foo.respond_to?(method) } }
  bm.report('respond_to_without_missing_nil?') { times.times { foo.respond_to_without_missing_nil?(method) } }
  bm.report('respond_to_without_missing_false?') { times.times { foo.respond_to_without_missing_false?(method) } }
  bm.report('Object#method') { times.times { (foo.method(method) rescue nil)&.source_location } }
end
```

MRI 2.4
```
                                        user     system      total        real
Set#include?                        1.470000   0.000000   1.470000 (  1.476836)
Object#respond_to?                  1.410000   0.000000   1.410000 (  1.413980)
respond_to_without_missing_nil?     3.440000   0.000000   3.440000 (  3.447178)
respond_to_without_missing_false?   3.410000   0.000000   3.410000 (  3.407693)
Object#method                       3.610000   0.010000   3.620000 (  3.626111)
```

JRuby 9.1.13.0
```
                                        user     system      total        real
Set#include?                        1.800000   0.000000   1.800000 (  1.253027)
Object#respond_to?                  1.470000   0.020000   1.490000 (  1.303870)
respond_to_without_missing_nil?     2.720000   0.010000   2.730000 (  2.513326)
respond_to_without_missing_false?   2.600000   0.000000   2.600000 (  2.390742)
Object#method                       3.650000   0.010000   3.660000 (  3.214350)
```

I'll file another pull request now.

PR #473, opened 2017-11-23
Drop the setter methods cache in favour of respond_to? (see #472)

chewi:
The cache was not being updated when new attributes were added, causing very strange behaviour. Any previously unseen attributes passed to new with `send_only_modified_attributes` enabled would be missing from the request parameters. This is because change tracking is only effective for setter methods.

Updating the cache turned out to be more expensive than not having a cache at all. Using `respond_to?` would be fastest but this breaks things as `respond_to_missing?` returns true for any assignment method.

If `method(:foo=).source_location` returns nil then this may indicate that the method is missing but this is also true for methods in native code.

A faster and more reliable approach is using a fiber-local variable to make `respond_to_missing?` return false while doing these checks. It's a tad ugly but it's the best I can come up with.

### DIFF
--- a/lib/her/model/attributes.rb
+++ b/lib/her/model/attributes.rb
@@ -55,7 +55,17 @@ module Her
 
       # @private
       def respond_to_missing?(method, include_private = false)
+        return false if Thread.current[:her_respond_to_missing]
         method.to_s =~ /[?=]$/ || @_her_attributes.include?(method) || super
+      end
+
+      def respond_to_without_missing?(method, include_private = false)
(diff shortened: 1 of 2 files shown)


chewi:
Rebased. I think this one is particularly important and it stands alone from my other remaining changes.

chewi:
I'm leaving the company soon and I'd like to get some (or all!) of these merged before I go. This one seems like a good start.

chewi:
One week to go. :stuck_out_tongue:

PR #479, opened 2017-12-22
Serialize ActiveModel instances that do not use Her

chewi:
Faraday does not serialize everything thrown at it and passing such instances currently just results in useless ugly strings.

`ActiveModel::Serialization` needs to be included for this to work.

### DIFF
--- a/lib/her/model/parse.rb
+++ b/lib/her/model/parse.rb
@@ -41,6 +41,17 @@ module Her
             end
           end
 
+          # Faraday does not serialize everything thrown at it and
+          # some parameters may still be complex objects by this point
+          # so attempt to serialize further where possible. This is
+          # useful for classes that use ActiveModel but not Her.
+          filtered_attributes.each do |key, value|
+            case value
+            when ActiveModel::Serialization
+              filtered_attributes[key] = value.serializable_hash
+            end
+          end
+
           if include_root_in_json?
             if json_api_format?
               { included_root_element => [filtered_attributes] }
--- a/spec/model/parse_spec.rb
+++ b/spec/model/parse_spec.rb
@@ -342,4 +342,30 @@ describe Her::Model::Parse do
       expect(user.to_params).to eql(user: { first_name: "Someone" })
     end
   end
+
+  context 'when passed a non-Her ActiveModel instance' do
+    before do
+      klass = Class.new do
+        include ActiveModel::Serialization
+
+        def attributes
+          { name: nil }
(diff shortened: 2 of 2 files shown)


lib/her/model/parse.rb
```diff
@@ -41,6 +41,17 @@ def to_params(attributes, changes={})
             end
           end
 
+          # Faraday does not serialize everything thrown at it and
```
zacharywelch: Let's instead make this part of the [initial assignment](https://github.com/remiprev/her/pull/479/files#diff-9f46c6432f36a5642681207c26cf15f8R35). Also fine removing the comment.

```ruby
filtered_attributes =
  attributes.dup.each_with_object({}) do |(key, value), memo|
    value = value.serializable_hash if value.is_a? ActiveModel::Serialization
    memo[key] = value
  end.symbolize_keys
```

spec/model/parse_spec.rb
```diff
@@ -342,4 +342,30 @@ def to_params
...
+
+      @model = klass.new
+
+      Her::API.setup
+      spawn_model 'Foo::User'
+    end
+
+    it 'serializes the instance in `to_params`' do
+      attributes = { model: @model }
+      expect(Foo::User.to_params(attributes)).to eq(model: { name: 'foo' })
```
zacharywelch: Let's test the public version of `#to_params` since `.to_params` is marked private.

```ruby
    it 'serializes the instance in `to_params`' do
      attributes = { model: @model }
      user = Foo::User.new(attributes)
      expect(user.to_params).to eq(model: { name: 'foo' })
    end
```

lib/her/model/parse.rb
chewi: Although I think this should work and it does pass in the context of this branch, I'm getting 3 failures when merged with my other branches. Investigating.

chewi: Never mind, turned out I'd resolved a conflict the wrong way.

chewi:
This addresses the changes requested. I changed your suggestion slightly.

* No `dup` is necessary.
* More efficient to `to_sym` on insertion than `symbolize_keys` afterwards.
* Need to `symbolize_keys` the `serializable_hash` if you want to keep everything symbolized.
* My other branch indirectly includes `ActiveModel::Serialization` in `Her::Model` and others may do this already so I've used a case statement to avoid calling `serializable_hash` on `Her::Model` instances.

Issue #474, opened 2017-11-29
Associations using the class_name option silently fail if the class fails to load


blazingkin:
Example of the issue
======
Suppose you have a her model that looks something like this
```
class Foo
    include Her::Model
    has_many :bars, class_name: 'Bar'
end
```
where 'Bar' is some other model.

Description of the issue
======
This model actually has a really subtle bug that is hard to notice.

Because the constant Bar is never referenced in this class, [rails never autoloads it](http://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoloading-algorithms). This means that if Foo is loaded before Bar, the reference to Bar is undefined and an error should be thrown.

Except - Her uses the following to find the class from the 'class_name' opt
` her_sibling_class(name) || name.constantize rescue nil`
i.e. if the class can't be found, it returns nil.

So now, instead of the code blowing up, if Foo loads before Bar the model behaves as if it was declared this way -
```
class Foo
    include Her::Model
    has_many :bars
end
```
Which is clearly not what was intended.

Why it matters
======
My team has had tests failing on our build server apparently randomly for about 5 weeks due to this issue. We could not recreate the failure locally, and the error message made no sense.

The error that actually gets thrown is that bars will be of type HashWithIndifferentAccess rather than the intended type of Bar. The error will certainly be thrown later in the program, so it will be difficult to trace back to the root cause.

This is compounded by the fact that since this is a load order bug, it will not happen all the time and will happen differently in different environments.

The fix
======
The correct model, without the bug is
```
require 'bar'
class Foo
    include Her::Model
    has_many :bars, class_name: 'Bar'
end
```

A stupidly simple fix.

What Her can do
======
Rather than just silently continuing if the class cannot be loaded, throw an error message. It would have saved us a lot of time and effort.

edtjones:
Thanks for the report @blazingkin - we could check Bar is defined when defining the association I guess. Fancy putting together a PR?

blazingkin:
Yeah, I can do that.

edtjones:
Perfect - thanks in advance!

zacharywelch:
Gonna go ahead and close this since #480 was merged 😺

thornfoot:
What is the best-practice for avoiding this exception in the first place?

blazingkin:
Hi @thornfoot!
Hope this didn't break anything in CAR

The simplest way to fix the error is to `require 'bar'`

[...]

PR #480, opened 2017-12-23
Raise an error when a class cannot be found

blazingkin:
Fixes issue #474.

Her now will raise an error rather than ignoring the class_name parameter if the class cannot be loaded.


blazingkin:
Hi @edtjones @remiprev, any chance someone could look at this. It's a super small change :)

edtjones:
Pinging @zacharywelch - dyou mind sanity checking, Zach?

edtjones:
thanks @zacharywelch :)

PR #483, opened 2018-01-29
 Prevent 'scope' from defining scope methods on the wrong Her models

smudge:
I've added tests to cover the cases I've encountered. Before my fix, you would see these new examples fail, because scope methods were being defined globally on `Her::Model::Relation`. (I also added one passing default_scope spec to test for an issue that I wanted to avoid introducing.)

To demonstrate the issue, take these two Her models:

```ruby
class B::User
  include Her::Model

  scope :foo, -> { where(bar: true) }
end

class A::User
  include Her::Model

  scope :foo, -> { where(bar: false) }
end
```

#### Before this fix:

```ruby
A::User.scoped.foo.params[:bar]
#=> false
B::User.scoped.foo.params[:bar]
#=> false -- OHNO!
```

Whoops! We can check, and the scope method definitions are actually the same:

```ruby
A::User.scoped.method(:foo).unbind ==  B::User.scoped.method(:foo).unbind
# => true
```

Note that the `scoped` is necessary here. It only affects _chained_ scoped, not scopes called directly on the class:

```ruby
A::User.foo.params[:bar]
#=> false
B::User.foo.params[:bar]
#=> true
A::User.method(:foo).unbind ==  B::User.method(:foo).unbind
# => false
```

#### After this fix:

```ruby
A::User.scoped.foo.params[:bar]
#=> false
B::User.scoped.foo.params[:bar]
#=> true
A::User.scoped.method(:foo).unbind ==  B::User.scoped.method(:foo).unbind
# => false
```

🎉

### DIFF
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -174,10 +174,8 @@ module Her
             instance_exec(*args, &code)
           end
 
-          # Add the scope method to the Relation class
-          Relation.instance_eval do
-            define_method(name) { |*args| instance_exec(*args, &code) }
-          end
+          # Add the scope method to the blank_relation
+          blank_relation.define_singleton_method(name) { |*args| instance_exec(*args, &code) }
         end
 
         # @private
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -267,9 +267,9 @@ module Her
           resource
         end
 
-        private
         # @private
         def blank_relation
+          @blank_relation ||= superclass.blank_relation.clone.tap { |r| r.parent = self } if superclass.respond_to?(:blank_relation)
           @blank_relation ||= Relation.new(self)
         end
       end
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -175,7 +175,7 @@ module Her
           end
 
           # Add the scope method to the blank_relation
-          blank_relation.define_singleton_method(name) { |*args| instance_exec(*args, &code) }
+          scoped.define_singleton_method(name) { |*args| instance_exec(*args, &code) }
         end
 
         # @private
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -174,7 +174,7 @@ module Her
             instance_exec(*args, &code)
           end
 
-          # Add the scope method to the blank_relation
+          # Add the scope method to the default/blank relation
           scoped.define_singleton_method(name) { |*args| instance_exec(*args, &code) }
         end
 
--- a/spec/model/relation_spec.rb
+++ b/spec/model/relation_spec.rb
@@ -79,8 +79,8 @@ describe Her::Model::Relation do
       end
 
       it "propagates the scopes through its children" do
-        @users = User.page(2)
-        expect(@users.length).to eq(2)
+        expect(User.page(2).length).to eq(2)
+        expect(User.scoped.page(2).length).to eq(2)
       end
     end
   end
--- a/spec/model/relation_spec.rb
+++ b/spec/model/relation_spec.rb
@@ -143,6 +143,7 @@ describe Her::Model::Relation do
         scope :bar, ->(v) { where(where: v) }
         scope :baz, -> { bar(6) }
       end
+      spawn_model 'Bar::User'
     end
 
     it "passes query parameters" do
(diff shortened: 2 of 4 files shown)


spec/model/relation_spec.rb
```diff
@@ -176,6 +209,27 @@ class User < Foo::Model; end
       end
     end
 
+    context 'when other scopes are present' do
```
zacharywelch: didn't notice a difference running these specs w/ the previous code. would be fine removing this block and sticking with the coverage you added above.

spec/model/relation_spec.rb
```diff
@@ -159,6 +159,39 @@ class User < Foo::Model; end
       @user = Foo::User.baz.first
       expect(@user.id).to eq(4)
     end
+
+    context 'when there is another model present' do
```
zacharywelch: Agree w/ the nature of these specs. Is it possible to combine them into one spec and narrow the coverage. How about?

```ruby
before do
  Her::API.setup url: "https://api.example.com" do |builder|
    builder.adapter :test do |stub|
      # ...
      stub.get("/users?where=6") { ok! [{ id: 4, fullname: "Tobias Fünke" }] }
      stub.get("/users?where=7") { ok! [{ id: 2, fullname: "Lindsay Fünke" }] }
      stub.get("/users") { ok! [{ id: 1, fullname: "Tobias Fünke" }] }
    end
  end

  spawn_model "Foo::User" do
    # ...
    scope :baz, -> { bar(6) }
  end

  spawn_model "Bar::User" do
    scope :baz, -> { where(where: 7) }
  end
end

# ...

it "does not share scope with other models" do
  expect(Bar::User.scoped).not_to respond_to(:foo, :bar)
  expect(Foo::User.scoped.baz.params[:where]).to eq(6)
  expect(Bar::User.scoped.baz.params[:where]).to eq(7)
end
```

lib/her/model/orm.rb
```diff
@@ -174,10 +174,8 @@ def scope(name, code)
             instance_exec(*args, &code)
           end
 
-          # Add the scope method to the Relation class
-          Relation.instance_eval do
-            define_method(name) { |*args| instance_exec(*args, &code) }
-          end
+          # Add the scope method to the default/blank relation
+          scoped.define_singleton_method(name) { |*args| instance_exec(*args, &code) }
```
zacharywelch: Noticed an issue with child classes no longer inheriting scopes. Can we make sure that's still available?

Here's a failing spec
```ruby
before do
  spawn_model "Foo::User" do
    scope :active, -> { where(status: "active") }
  end

  Foo::Admin = Class.new(Foo::User)
end

it "passes scopes to children" do
  expect(Foo::Admin.scoped).to respond_to(:active)
end
```

spec/model/relation_spec.rb
smudge: Combined. 👍

lib/her/model/orm.rb
smudge: There was already a spec for scope inheritance but it didn't cover the chained scope case -- updated that to support both.

lib/her/model/orm.rb
```diff
@@ -269,9 +267,9 @@ def build(attributes = {})
           resource
         end
 
-        private
         # @private
         def blank_relation
+          @blank_relation ||= superclass.blank_relation.clone.tap { |r| r.parent = self } if superclass.respond_to?(:blank_relation)
```
smudge: This now pulls the blank relation of of the superclass, similar to the way `default_scope` is inherited.

The one API change to `Relation` was to add a writer for `parent` so it could be updated from this context.

spec/model/relation_spec.rb
```diff
@@ -135,6 +135,7 @@ class User < Foo::Model; end
           stub.get("/users?what=4&where=3") { ok! [{ id: 3, fullname: "Maeby Fünke" }] }
           stub.get("/users?what=2") { ok! [{ id: 2, fullname: "Lindsay Fünke" }] }
           stub.get("/users?where=6") { ok! [{ id: 4, fullname: "Tobias Fünke" }] }
+          stub.get('/bar/users') { ok! [] }
```
smudge: `Bar::User.scoped#respond_to?` is set to do a `fetch` under the hood, which is why this stub was necessary. I gave the model its own `collection_path` to make sure this stub wouldn't interfere with the other model in this context.

smudge:
Alright, addressed feedback and made sure specs are green. Let me know if you have any questions or additional feedback on my changes to support inheritance. (Good catch on that!) 👍

Issue #488, opened 2018-03-15
Can't set `include_root_in_json` for child models in some cases


zoltankiss:
let's say we have a parent model:

```ruby
class HerParent
  include Her::Model
  include_root_in_json true
end
```

Let's further say I want to set the child model's `include_root_in_json` to be `false`:

```ruby
class HerChild < HerParent
  include_root_in_json false
end
```

Unfortunately [here](https://github.com/remiprev/her/blob/a29c883f2ed875908825e06b4f714f79d0116e41/lib/her/model/parse.rb#L202) the logic works like this:

```ruby
def include_root_in_json?
  @_her_include_root_in_json || (superclass.respond_to?(:include_root_in_json?) && superclass.include_root_in_json?)
end
```

So since I set `include_root_in_json`  to be `false` on the child model, and that is falsey, `a || b` returns `b` in ruby if `a` is falsey.

The correct fix would be something like this:

```ruby
def include_root_in_json?
  if @_her_include_root_in_json.nil?
    superclass.respond_to?(:include_root_in_json?) && superclass.include_root_in_json?
  end
  @_her_include_root_in_json
end
```

A complete reproduction script:

```ruby
require 'her'

class HerParent
  include Her::Model
  include_root_in_json true
end

class HerChild < HerParent
  include_root_in_json false
end

class HerChildWithClassMethod < HerParent
  def self.include_root_in_json?
    false
  end
end

puts "HerChild.include_root_in_json? actual:#{HerChild.include_root_in_json?} expected: false"
puts "HerChildWithClassMethod.include_root_in_json? actual:#{HerChildWithClassMethod.include_root_in_json?} expected: false"
```

edtjones:
good spot @zoltankiss - could you put together a PR and I'll merge?

zoltankiss:
yup!

PR #490, opened 2018-03-20
Child classes can't set some attributes to false.

zoltankiss:
Fixes: #488

cc @edtjones


edtjones:
Terrific - thanks for this @zoltankiss - will do a release and push to rubygems ASAP.

zoltankiss:
🎉!

PR #491, opened 2018-04-11
Refactor associations_spec

ssuttner:
The `spec/model/associations_spec.rb` file has complex test cases and I thought some refactoring would help increase its readability and understanding.

Here you'll find improvements such as:
* Rearrange contexts
* Remove instance variables in favor of `let`
* Prepare data for contexts in a more isolated manner
* Enhance overall readability

### DIFF
--- a/spec/model/associations_spec.rb
+++ b/spec/model/associations_spec.rb
@@ -8,10 +8,21 @@ describe Her::Model::Associations do
 
     context "single has_many association" do
       before { Foo::User.has_many :comments }
+      subject { super()[:has_many] }
 
       describe "[:has_many]" do
-        subject { super()[:has_many] }
-        it { is_expected.to eql [{ name: :comments, data_key: :comments, default: [], class_name: "Comment", path: "/comments", inverse_of: nil }] }
+        let(:association) do
+          {
+            name: :comments,
+            data_key: :comments,
+            default: [],
+            class_name: "Comment",
+            path: "/comments",
+            inverse_of: nil
+          }
+        end
+
+        it { is_expected.to eql [association] }
       end
     end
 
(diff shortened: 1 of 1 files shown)


spec/model/associations_spec.rb
```diff
@@ -8,10 +8,21 @@
 
     context "single has_many association" do
       before { Foo::User.has_many :comments }
+      subject { super()[:has_many] }
```
zacharywelch: while we're refactoring think we should use an explicit subject instead of calling `super()`?

```ruby
context "setting associations without options" do
  subject(:associations) { User.associations }
  # ...
  context "single has_many association" do
    subject { associations[:has_many] }
  end
```

spec/model/associations_spec.rb
```diff
@@ -2,6 +2,71 @@
 require File.join(File.dirname(__FILE__), "../spec_helper.rb")
 
 describe Her::Model::Associations do
+
+  context "building and creating association data" do
```
zacharywelch: Let's keep the `setting associations` specs up top. It's useful to first understand how association definitions are stored on the model before diving into how they influence build & creation.

spec/model/associations_spec.rb
```diff
@@ -546,6 +567,21 @@
     end
 
     context "when foreign_key is nil" do
+      before do
+        Her::API.setup url: "https://api.example.com" do |builder|
```
zacharywelch: still needed?

spec/model/associations_spec.rb
```diff
@@ -536,6 +550,13 @@
     end
 
     context "when resource is new" do
+      before(:context) do
```
zacharywelch: still needed?

spec/model/associations_spec.rb
ssuttner: I took it out :)

spec/model/associations_spec.rb
ssuttner: I took it out :)

ssuttner:
@zacharywelch fixes are done.
Thanks for the review!
Let me know if you find anything else :)

PR #493, opened 2018-04-13
Fix allow ':class' parameter

skswhwo:


zacharywelch:
Hi @skswhwo thanks for your PR 😺Do you mind including a brief description of the issue? If you can give an example or test case that would also help us tremendously.

If it helps clarify things, `reserved` is meant to correspond to the predefined primary key and associations on the model.

Thanks!

PR #496, opened 2018-04-13
Rubocop fixes

ssuttner:
Hello!

I'm creating this PR to take `her` one step closer to the [ruby-style-guide](https://github.com/bbatsov/ruby-style-guide) by the use of `rubocop`.

I tried to stick as much as possible to the defaults:
1) Created a few rules on `.rubocop.yml`, which are just proposed by me and open to discussion.
2) Had `rubocop` create a `.rubocop_todo.yml` with rules to skip a lot of offenses I was not sure about what this team thinks of them.

My idea here is to talk about each of the rules and decide what we want `her` to follow; would love to provide my input on them. I'm open to suggestions and I'm willing to make as many modifications as needed.
My vision behind this PR is to help on the objective of configuring `CodeClimate` and make PR reviews easier.

P.S. I know the list of commits is huge, I only did that to make the review easier. I can squash if they don't add value as they are now.

Thanks!


ssuttner:
@zacharywelch Travis let me know I need to take care of the old ruby versions. 
I'll take care of that later. Hope this serves as a good starting point.

ssuttner:
Here is a set of fixes for `rubocop` rules that apply to all ruby versions. Some others are pending, waiting for discussion -- mostly because fixes behave differently for each ruby version that's supported.

edtjones:
@ssutner thanks for getting into this. Great stuff.

.rubocop_todo.yml
```diff
@@ -43,18 +43,6 @@ Layout/EmptyLineBetweenDefs:
   Exclude:
     - 'spec/model/orm_spec.rb'
 
-# Offense count: 6
-# Cop supports --auto-correct.
-# Configuration parameters: EnforcedStyle.
-# SupportedStyles: empty_lines, empty_lines_except_namespace, empty_lines_special, no_empty_lines, beginning_only, ending_only
```
zacharywelch: let's opt for `beginning_only`

.rubocop_todo.yml
```diff
@@ -65,13 +65,6 @@ Lint/AmbiguousOperator:
   Exclude:
     - 'lib/her/model/attributes.rb'
 
-# Offense count: 2
```
zacharywelch: let's skip this cop

lib/her/api.rb
```diff
@@ -72,7 +72,7 @@ def setup(opts = {}, &blk)
       opts[:url] = opts.delete(:base_uri) if opts.include?(:base_uri) # Support legacy :base_uri option
       @options = opts
 
-      faraday_options = @options.reject { |key, value| !FARADAY_OPTIONS.include?(key.to_sym) }
+      faraday_options = @options.reject { |key, _value| !FARADAY_OPTIONS.include?(key.to_sym) }
```
zacharywelch: let's turn these into underscores instead

lib/her/model/associations.rb
```diff
@@ -26,12 +26,12 @@ def get_association(association_name)
...
       module ClassMethods
-        # Return @_her_associations, lazily initialized with copy of the
-        # superclass' her_associations, or an empty hash.
+        # Return @associations, lazily initialized with copy of the
+        # superclass' associations, or an empty hash.
         #
         # @private
         def associations
-          @_her_associations ||= begin
+          @associations ||= begin
```
zacharywelch: I prefer this style as well but don't know what the effects would be or the original reason for prefixing these instance variables with `_her.  For example would we run into a situation [like @attributes](https://github.com/remiprev/her/pull/497) where another class defines `@associations` and we have a naming conflict? Any thoughts @edtjones?

spec/model/callbacks_spec.rb
```diff
@@ -21,10 +21,12 @@
 
     context "when using a symbol callback" do
       before do
-        class Foo::User
-          before_save :alter_name
-          def alter_name
-            name.upcase!
+        module Foo
```
zacharywelch: agree with keeping this rule and replacing the class definitions with [spawn_model](https://github.com/remiprev/her/blob/e2605655915129c7a7664b6518d64aee25c4827e/spec/model/associations/association_proxy_spec.rb#L15) instead.

.rubocop_todo.yml
```diff
@@ -141,14 +141,6 @@ Style/ExpandPathArguments:
 Style/FrozenStringLiteralComment:
   Enabled: false
 
-# Offense count: 3
```
zacharywelch: let's skip this

lib/her/model/introspection.rb
```diff
@@ -50,7 +50,11 @@ def her_nearby_class(name)
         def her_sibling_class(name)
           return unless (mod = her_containing_module)
           @_her_sibling_class ||= Hash.new { Hash.new }
-          @_her_sibling_class[mod][name] ||= "#{mod.name}::#{name}".constantize rescue nil
+          @_her_sibling_class[mod][name] ||= begin
```
zacharywelch: let's skip

.rubocop_todo.yml
```diff
@@ -248,14 +248,6 @@ Style/TernaryParentheses:
     - 'lib/her/model/http.rb'
     - 'lib/her/model/orm.rb'
 
-# Offense count: 6
```
zacharywelch: let's skip

.rubocop_todo.yml
```diff
@@ -248,14 +248,6 @@ Style/TernaryParentheses:
...
-# Cop supports --auto-correct.
-# Configuration parameters: EnforcedStyle, MinSize, WordRegex.
-# SupportedStyles: percent, brackets
-Style/WordArray:
```
zacharywelch: let's skip

.rubocop_todo.yml
```diff
@@ -0,0 +1,247 @@
...
+# SupportedStyles: single_quotes, double_quotes
+Style/StringLiteralsInInterpolation:
+  Exclude:
+    - 'lib/her/model/introspection.rb'
+
+# Offense count: 6
+# Cop supports --auto-correct.
+# Configuration parameters: MinSize.
+# SupportedStyles: percent, brackets
+Style/SymbolArray:
```
zacharywelch: let's skip

zacharywelch:
@ssuttner thanks so much for these improvements! 🎩🙌

PR #507, opened 2018-06-29
Add save_existing! to Her::Model::ORM

ikuto0608:
Similar to User.save_existing(), except that ResourceInvalid is raised if the save fails.

### DIFF
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -223,6 +223,13 @@ module Her
           resource
         end
 
+        # Similar to User.save_existing(), except that ResourceInvalid is raised if the save fails
+        def save_existing!(id, params)
+          resource = new(params.merge(primary_key => id))
+          resource.save!
+          resource
+        end
+
         # Destroy an existing resource
         #
         # @example
--- a/lib/her/model/orm.rb
+++ b/lib/her/model/orm.rb
@@ -218,12 +218,12 @@ module Her
         #   @user = User.save_existing(1, { :fullname => "Tobias Fünke" })
         #   # Called via PUT "/users/1"
         def save_existing(id, params)
-          resource = new(params.merge(primary_key => id))
-          resource.save
-          resource
+          save_existing!(id, params)
+        rescue Her::Errors::ResourceInvalid => e
+          e.resource
         end
 
-        # Similar to User.save_existing(), except that ResourceInvalid is raised if the save fails
+        # Similar to .save_existing but raises ResourceInvalid if save fails
         def save_existing!(id, params)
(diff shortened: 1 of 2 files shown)


ikuto0608:
Hi @edtjones 

There was the situation I didn’t want to use `save/save!` instead.

ex)
```
# External api model with her
class External::User
  attr_reader :hoge
  attr_accessor :fuga
end

user = External::User.find(1)
user.fuga = 'anything' # :fuga is allowed to change

# failed
user.save # -> failed because it sent a PUT request with all elements including `:hoge`,
          # and the external api checks whether the params includes only rewritable params or not,
          # and if the params has read only item, it's not gonna just ignore but `422 UNPROCESSABLE ENTITY`.
          # So I needed to send only the params which I wanted to modify.

# success
External::User.save_existing(user.id, { fuga: 'anything' })
```

Thanks!

lib/her/model/orm.rb
```diff
@@ -223,6 +223,13 @@ def save_existing(id, params)
           resource
         end
 
+        # Similar to User.save_existing(), except that ResourceInvalid is raised if the save fails
```
zacharywelch: Let's go ahead and refactor `save_existing` to use your method

```ruby
def save_existing(id, params)
  save_existing!(id, params)
rescue Her::Errors::ResourceInvalid => e
  e.resource
end
```

zacharywelch: `Similar to .save_existing but raises ResourceInvalid if save fails`

spec/model/orm_spec.rb
```diff
@@ -423,6 +424,15 @@ def friends
       expect(@user.fullname).to eq("Lindsay Fünke")
     end
 
+    it "handle resource update through the .update class method" do
```
zacharywelch:
```ruby
it "handle resource update through .save_existing!" do
```

spec/model/orm_spec.rb
```diff
@@ -423,6 +424,15 @@ def friends
...
+      @user = Foo::User.save_existing!(1, fullname: "Lindsay Fünke")
+      expect(@user.fullname).to eq("Lindsay Fünke")
+    end
+
+    it "raises ResourceInvalid when #save_existing! gets errors" do
```
zacharywelch:
```ruby
it "raises ResourceInvalid when .save_existing! has errors" do
```

ikuto-rise:
Thanks @zacharywelch for doing code review, I updated the code!

ikuto0608:
Thanks @zacharywelch for doing code review, I updated the code!

PR #511, opened 2018-10-17
fix path option on belongs_to association

### LINKED ISSUE #506: Belongs_to options are not used when building request
I did some digging and could be mistaken, but it seems to me that the path option in the belongs_to association method should override the path that the association is found at. 
The line that builds the path is found [here](https://github.com/remiprev/her/blob/master/lib/her/model/associations/belongs_to_association.rb#L83), but does not seem to use the path option in any way [here](https://github.com/remiprev/her/blob/master/lib/her/model/associations/belongs_to_association.rb#L14). 

I am attempting to set up an association as follows:
```ruby
belongs_to :something, 
  foreign_key: 'something_internal_symbol', 
  path: 'something/is/:internal_symbol', 
  primary_key: 'internal_symbol'
```

but the path that is requested is always something/:id, which is of course that object's primary key.
This is because the path is built by asking the association class, without first checking if the path option is set. 

Do folks agree that should be checked first?

Additionally, other options, such as the primary key specification, are also not used when building request params. 

Perhaps I am misunderstanding the usage of the options upon setup?

zacharywelch:
fixes #506 

`path` option was being ignored on `belongs_to` associations and replaced with the default. According to the [documentation](https://www.rubydoc.info/github/remiprev/her/Her%2FModel%2FAssociations%2FClassMethods:belongs_to) 

> Options Hash (opts):
> :path (Path) — The relative path where to fetch the data (defaults to `/class_name.pluralize/id`)

#### Example of the behavior

```ruby
class User
  include Her::Model
  
  belongs_to :company, path: "/organizations/:id", 
                       foreign_key: :organization_id, 
                       data_key: :organization
end


class Company
  include Her::Model
end
```

**Before**

```
> user = User.find(1) 
# GET "/users/1", response is:
# {
#   "id": 1,
#   "name": "George Michael Bluth",
#   "organization_id": 2
# }

> user.company # goes to wrong path
# GET "/companies/2"
```

**After**

```
> user = User.find(1) 
# GET "/users/1", response is:
# {
#   "id": 1,
#   "name": "George Michael Bluth",
#   "organization_id": 2
# }

> user.company # goes to correct path
# GET "/organizations/2"
```

### DIFF
--- a/lib/her/model/associations/belongs_to_association.rb
+++ b/lib/her/model/associations/belongs_to_association.rb
@@ -80,7 +80,7 @@ module Her
           return @parent.attributes[@name] unless @params.any? || @parent.attributes[@name].blank?
 
           path_params = @parent.attributes.merge(@params.merge(@klass.primary_key => foreign_key_value))
-          path = build_association_path -> { @klass.build_request_path(path_params) }
+          path = build_association_path -> { @klass.build_request_path(@opts[:path], path_params) }
           @klass.get_resource(path, @params).tap do |result|
             @cached_result = result if @params.blank?
           end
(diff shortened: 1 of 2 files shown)


edtjones:
Thanks as ever for being a hero @zacharywelch 👍

seanbjornsson:
Thanks for the fix @zacharywelch !

PR #514, opened 2018-10-20
belongs_to association should respect association's collection_path

### LINKED ISSUE #513: belongs_to associations no longer respect collection_paths
The change in version 1.0.3 caused a regression in the handling of paths for the belongs_to association. It accepts an optional path now, but the default value for that option ignores the associated class's collection path. Simple broken example:
```
class Company
  include Her::Model
  include_root_in_json true
  collection_path 'my_companies'
end

class Admin
  include Her::Model
  include_root_in_json true
  belongs_to :company
end
```
in order to make the same thing work with 1.0.3, you would have to write `belongs_to :company, path: "#{Company.collection_path}/:company_id"` in the admin class.

One simple solution is to not provide a default value for the path argument.

keegangroth:
fixes #513

### DIFF
--- a/lib/her/model/associations.rb
+++ b/lib/her/model/associations.rb
@@ -116,7 +116,7 @@ module Her
         # @param [Hash] opts Options
         # @option opts [String] :class_name The name of the class to map objects to
         # @option opts [Symbol] :data_key The attribute where the data is stored
-        # @option opts [Path] :path The relative path where to fetch the data (defaults to the target class's path)
+        # @option opts [Path] :path The relative path where to fetch the data
         # @option opts [Symbol] :foreign_key The foreign key used to build the `:id` part of the path (defaults to `{name}_id`)
         #
         # @example


spec/model/associations_spec.rb
```diff
@@ -112,24 +112,39 @@
             data_key: :organization,
             default: nil,
             class_name: "Organization",
-            foreign_key: "organization_id",
-            path: "/organizations/:id"
```
zacharywelch: We need each association in the list of `<HerModel>.associations` to include a path. The documentation states this will default to `/{class_name}.pluralize/{id}` for `belongs_to` though I agree it should be `resource_path` or `collection_path` depending on the type of association.

keegangroth: ok I guess that means extracting and generalizing the path building logic from fetch. I'll give it a try when I get a chance

zacharywelch: @keegangroth after further thought I'm more inclined to accept a default `path` of `nil` on `belongs_to` 👍  Long term we might want to refactor `.associations` to resemble something like [ActiveRecord::Reflection](https://api.rubyonrails.org/classes/ActiveRecord/Reflection/ClassMethods.html) 😅 

Until then, defaulting to `nil` would have the following, expected behavior on belongs_to associations: use `path` if provided, otherwise `build_request_path` would [default](https://github.com/remiprev/her/blob/fd192875777f11135ec2188a4a9b8346226b9561/lib/her/model/paths.rb#L96) to `resource_path`. We'd also need to update the [docs](https://github.com/remiprev/her/blob/master/lib/her/model/associations.rb#L119) for belongs_to (maybe remove the _defaults to..._ language)

@edtjones see any issues here?

keegangroth: sounds good, I was not seeing a good way to get at the relevant information where the option defaults were being set. Will have a go at updating the docs.

zacharywelch:
thanks @keegangroth 🎩 🎉

PR #516, opened 2018-11-26
correctly dup objects

### LINKED ISSUE #355: `dup` and `clone` doesn't work on Her::Models
I tried to run a `dup` or `clone` on instance of **Her::Model** but it doesn't work and attributes are shared between copies of that instance. I believe it should work differently and the internal state of object should be copied on `clone`.

```
$ irb
2.0.0-p247 :001 > require 'her'
 => true
2.0.0-p247 :002 > class HerModel; include Her::Model;  end
 => HerModel
2.0.0-p247 :003 > i1 = HerModel.new(foo: 42)
 => #<HerModel(her_models) foo=42>
2.0.0-p247 :004 > i2 = i1.clone
 => #<HerModel(her_models) foo=42>
2.0.0-p247 :005 > i2.foo = 7
 => 7
2.0.0-p247 :006 > i2
 => #<HerModel(her_models) foo=7>
2.0.0-p247 :007 > i1
 => #<HerModel(her_models) foo=7>
2.0.0-p247 :08 > i1.object_id
 => 70162419040340
2.0.0-p247 :09 > i2.object_id
 => 70162431153300
```

KaylaGallatin:
Attempting to fix #355 

After calling `dup` on an object, changes made to either object would be applied to both objects. This fixes that issue so that both objects are independent of one another.

### DIFF
--- a/lib/her/model/base.rb
+++ b/lib/her/model/base.rb
@@ -28,6 +28,17 @@ module Her
       def singularized_resource_name
         self.class.name.split('::').last.tableize.singularize
       end
+
+      # @private
+      def initialize_dup(other)
+        @_her_attributes = @_her_attributes.dup
+        @_her_attributes[self.class.primary_key] = nil
+
+        @new_record  = true
+        @destroyed   = false
+
+        super
+      end
     end
   end
 end
--- a/lib/her/model/base.rb
+++ b/lib/her/model/base.rb
@@ -28,17 +28,6 @@ module Her
       def singularized_resource_name
         self.class.name.split('::').last.tableize.singularize
       end
-
-      # @private
-      def initialize_dup(other)
-        @_her_attributes = @_her_attributes.dup
-        @_her_attributes[self.class.primary_key] = nil
-
-        @new_record  = true
-        @destroyed   = false
-
-        super
-      end
     end
   end
 end
--- a/lib/her/model/associations.rb
+++ b/lib/her/model/associations.rb
@@ -25,6 +25,15 @@ module Her
         send(association_name) if has_association?(association_name)
       end
 
+      # @private
+      def clear_associations
+        self.class.association_names.each do |association|
+          cached_name = :"@_her_association_#{association}"
+          instance_variable_set(cached_name, nil) if instance_variable_defined?(cached_name)
+          attributes.delete(association)
+        end
+      end
+
       module ClassMethods
         # Return @_her_associations, lazily initialized with copy of the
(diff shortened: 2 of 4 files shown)


lib/her/model/base.rb
```diff
@@ -28,6 +28,17 @@ def [](attribute_name)
       def singularized_resource_name
         self.class.name.split('::').last.tableize.singularize
       end
+
+      # @private
+      def initialize_dup(other)
+        @_her_attributes = @_her_attributes.dup
+        @_her_attributes[self.class.primary_key] = nil
+
+        @new_record  = true
```
zacharywelch: we can remove this since [new_record?](https://github.com/remiprev/her/blob/master/lib/her/model/orm.rb#L13) doesn't rely on an instance variable 😺

lib/her/model/base.rb
```diff
@@ -28,6 +28,17 @@ def [](attribute_name)
       def singularized_resource_name
         self.class.name.split('::').last.tableize.singularize
       end
+
+      # @private
+      def initialize_dup(other)
+        @_her_attributes = @_her_attributes.dup
+        @_her_attributes[self.class.primary_key] = nil
+
```
zacharywelch: To stick w/ [ActiveRecord](https://api.rubyonrails.org/classes/ActiveRecord/Core.html#method-i-dup) behavior, we'll also need to clear or assign `nil` to any association caches. 

Associations which aren't included in the initial request are stored in separate [instance variables](https://github.com/remiprev/her/blob/master/lib/her/model/associations/has_many_association.rb#L20) following a `@_her_association_<name>` convention. 

Associations which are part of the initial request are [merged](https://github.com/remiprev/her/blob/master/lib/her/model/attributes.rb#L88) w/ `@_her_attributes`.

lib/her/model/base.rb
```diff
@@ -28,6 +28,17 @@ def [](attribute_name)
       def singularized_resource_name
         self.class.name.split('::').last.tableize.singularize
       end
+
+      # @private
+      def initialize_dup(other)
```
zacharywelch: can we go ahead and move this to `attributes.rb` below [initialize](https://github.com/remiprev/her/blob/master/lib/her/model/attributes.rb#L26)?

Skippz88 (approved):

lib/her/model/base.rb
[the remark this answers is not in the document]
Skippz88: should we also run_callbacks :initialize here?

PR #530, opened 2019-10-13
Remove unused variable from Model::Http#custom method

vaihtovirta:
The `opts` variable inside `Http#custom_` was introduced in [88973d7](https://github.com/remiprev/her/commit/88973d7be54de7e7b4c8348609cb567e0ce4f0db#diff-8748e6bfaf520753b590aabc01e77c7aR76) and was overlooked during removing caching functionality in [942a4abd](https://github.com/remiprev/her/commit/942a4abd392ca85e895e8348fdad828b4c37f028#diff-8748e6bfaf520753b590aabc01e77c7aL81).
I got `assigned but unused variable - opts` warning in one of my projects and after checking, discovered that the variable could be indeed removed.

### DIFF
--- a/lib/her/model/http.rb
+++ b/lib/her/model/http.rb
@@ -97,7 +97,6 @@ module Her
 
             def custom_#{method}(*paths)
               metaclass = (class << self; self; end)
-              opts = paths.last.is_a?(Hash) ? paths.pop : Hash.new
 
               paths.each do |path|
                 metaclass.send(:define_method, path) do |*params|
--- a/lib/her/model/http.rb
+++ b/lib/her/model/http.rb
@@ -98,6 +98,12 @@ module Her
             def custom_#{method}(*paths)
               metaclass = (class << self; self; end)
 
+              # TODO: Remove this check after January 2020
+              if paths.last.is_a?(Hash)
+                warn("[DEPRECATION] options for custom request methods are deprecated and will be removed on or after January 2020.")
+                paths.pop
+              end
+
               paths.each do |path|
                 metaclass.send(:define_method, path) do |*params|
                   params = params.first || Hash.new
--- a/spec/model/http_spec.rb
+++ b/spec/model/http_spec.rb
@@ -158,18 +158,16 @@ describe Her::Model::HTTP do
     subject { Foo::User }
 
     describe :custom_get do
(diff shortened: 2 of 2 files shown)


lib/her/model/http.rb
```diff
@@ -97,7 +97,6 @@ def #{method}_resource(path, params={})
 
             def custom_#{method}(*paths)
               metaclass = (class << self; self; end)
-              opts = paths.last.is_a?(Hash) ? paths.pop : Hash.new
```
zacharywelch: Nice find 👍 Only risk may be if someone passed in options expecting them to be used, and then [L103](https://github.com/remiprev/her/blob/master/lib/her/model/http.rb#L103) failed. Another option could be 
```ruby
paths.pop if paths.last.is_a?(Hash)
```
That said, `custom_` options [doesn't](https://github.com/remiprev/her/blob/master/lib/her/model/http.rb#L14) [appear](https://github.com/remiprev/her#custom-requests) to be documented so perhaps it is ok to remove this line altogether. @edtjones any thoughts?

edtjones: I would prefer something non-breaking 🙂 - I have a vague recollection of using this method somewhere.

edtjones: (But we should document it if we are keeping it, and add some tests 🙂)

vaihtovirta: Thanks for your responses.
Since these options aren't really used further in the method, should we also put some deprecation warning along with the line mentioned above? 
Something like:
```ruby
warn "[DEPRECATION] options for custom request methods are deprecated and will be removed in the next release."
paths.pop if paths.last.is_a?(Hash)
```

zacharywelch: That'd be awesome 😎 Can we issue the warning only if opts are passed in? And tweak the message to something like?

 ```
[DEPRECATION] options for custom request methods are deprecated and will be removed on or after January 2020.
```

zacharywelch:
Made a small tweak to the spec and squashed before merging to [master](https://github.com/remiprev/her/commits/master). Thanks for your help w/ this @vaihtovirta! 🎩👍

edtjones:
and also big thanks from me @vaihtovirta! ❤️

PR #539, opened 2020-09-10
Fix infinite loop when parsing nested associations

gordysc:
It appears [with this commit](https://github.com/remi/her/commit/69406cffae856922210f069a083df99a44aa5247?branch=69406cffae856922210f069a083df99a44aa5247&diff=split#diff-9f46c6432f36a5642681207c26cf15f8R57) a potential infinite loop was introduced if models defined both has_many/belongs_to.  

To mitigate this, I've added a straightforward check for `belongs_to` to no-op if the association is already defined on the resulting object.  Also, sorry for the previous PR!  Moved this under my company's organization for easy access to the rest of my team.

### DIFF
--- a/lib/her/model/parse.rb
+++ b/lib/her/model/parse.rb
@@ -61,15 +61,23 @@ module Her
 
         # @private
         def embeded_params(attributes)
-          associations.values.flatten.each_with_object({}) do |definition, hash|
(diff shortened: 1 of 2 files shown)


gabrielrdrguez:
Somebody please give a medal to this man! (and merge this PR :stuck_out_tongue: )

PR #540, opened 2021-03-23
has_many_association: Deal with empty 204 status code

pedrolucasp:
If you register a relationship as a `has_many` and you this resource
endpoint return a 204 we should handle this gracefully, by returning a
empty array. This patch leverages the default return type defined on the
relationship (`@opts[:default]`) to construct which sort of response it
would expect (`get_collection` or a regular `get`).

The default parser, is dumb and therefore cannot distinguish if when it
receives a 204 it should transform that into a hash or an array, and
always convert the response into a empty hash. So, on the
attributes.rb method called `instantiate_collection` when we stumble
upon a empty hash we always convert into an empty array.


pedrolucasp:
Hello, just a quick ping on this PR for review @remi @zacharywelch :)

PR #548, opened 2023-04-13
Update dependencies

kates:
Updated dependencies to use the latest and greatest as of this moment.


kates:
Closing the PR accidentally made here instead of my own fork.

